PR #14, opened 2021-02-08
Annotation of ISA-tab data from TSV

### LINKED ISSUE #3: Add "cubi-tk isa-tab annotate" command
The aim is to be able to support the following use cases:

- annotate sources with their HLA types stored in a TSV file
- annotate samples with QC statistics such as read counts etc.

The input would be:

- folder with ISA-tab files
- path to annotation TSV file
- which assay/study to annotate
- name of the output column

The output would be

- folder with ISA-tab files where the study/assay files are updated to reflect the annotations

For example (pseudo ISA-tab)

input ISA-tab has

Source Name | Sample Name
------------|------------
SampleA | SampleA-N1
SampleB | SampleB-N1
SampleC | SampleC-N1

Annotation has

Source Name | HLA Type
------------|---------
SampleA | A:01:01;A01:02
SampleC | A:03:01;A03:02

Result is

Source Name | HLA Type | Sample Name
------------|----------|------------
SampleA | A:01:01;A01:02 | SampleA-N1
SampleB | | SampleB-N1
SampleC | A:03:01;A03:02 | SampleC-N1

We can then also create a `cubi-tk sodar isa-tab-annotate` that does it to the ISA-tab of a SODAR project.

cc @ericblanc20

holtgrewe:
closes #3


eudesbarbosa:
Implemented in #25

PR #28, opened 2021-02-15
Fix configuration for single_cell_rnaseq ISAtab template

mikkonie:
The configuration was pointing to `isatab-generic`, seems to me it should be `isatab-single_cell_rnaseq` instead.


holtgrewe:
lgtm

PR #34, opened 2021-02-22
Feature/select assay in pull raw data

ericblanc20:
# Assay selection in `pull-raw-data`

- User can select a specific assay using the optional argument `--assay ASSAY` for the `cubi-tk sodar pull-raw-data` and `cubi-tk snappy pull-raw-data` commands. The behaviour for projects with a single assay is unchanged.
- The naming of arguments has been enforced for classes defined in `cubi_tk/sodar/models.py` and `cubi_tk/snappy/models.py`. This fixes `cattr.structure` problems when building new objects with optional arguments. A related problem was described [here](https://github.com/python-attrs/attrs/issues/93).
- The contents of the Isa-Tab assay `Library Name` column is a default when the `Folder Name` column is missing from that file.

### DIFF
--- a/cubi_tk/snappy/models.py
+++ b/cubi_tk/snappy/models.py
@@ -9,7 +9,7 @@ from logzero import logger
 import yaml
 
 
-@attr.s(frozen=True, auto_attribs=True)
+@attr.s(frozen=True, auto_attribs=True, kw_only=True)
 class SearchPattern:
     """Represent an entry in the ``search_patterns`` list."""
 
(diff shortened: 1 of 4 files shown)


eudesbarbosa (changes requested):
I would suggest keep the variable names consistent: instead of `assay`, use `assay_uuid` - same as `project_uuid`.

ericblanc20:
OK, but `--assay` is used in `lz_create.py`, and it is only place that I could see where the assay uuid was used on the command line. So this is the reason why I used it rather than `--assay_uuid`. Perhaps we should change both?

eudesbarbosa:
I think the argument name is fine like `--assay`. I would just change the name of the internal destiny variant.

holtgrewe:
Just one comment, we should generally use [GNU style arguments](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html). `argparse` is taking care of most, but please use `--kebap-case` (or short `-k`) for arguments.

holtgrewe:
Also, please make sure to use squash merge for such small changes.

Otherwise: nice work, it's the sum of such small improvements that make a big leap! :+1:

Issue #27, opened 2021-02-12
Provide a way to install without pandas


holtgrewe:
**Describe the bug**
Pandas has many transient dependencies that are really unnecessary for our use case.
This is particularly important when using `cubi_tk` as a library from other code.
We should get rid of it as a dependency as it is used in few places only anyway or provide a `cubi-tk-minimal` version similar to how snakemake does it that disables some features but is independent of pandas.

mkuhring:
The impact on cubi-tk annotate would be minimal. I will take care of
removing it from there completely.

Manuel Holtgrewe <notifications@github.com> schrieb am Fr., 12. Feb. 2021,
15:06:

PR #38, opened 2021-03-16
Add isa-tab annotate (#3)

mkuhring:
Annotate sub-command is ready for use now and complete with tests and documentation. Study and assay selection is now handled by indicating filenames which are used directly in the annotate SheetUpdateVisitor. Only change I need to do to the isa_support models is the removal of the limit to one study/assay via exception. The isa_support models do iterate all studies and assays. So this exception seems to be not necessary anymore, anyways. No test complaint about it. Pandoc dependency is also removed (contributing to #27). Will squash if accepted.


holtgrewe:
@eudesbarbosa please also review

mkuhring:
Converted to draft, trying to fix codacy issues.

mkuhring:
Fixed Codacy issues. It's still complaining a bit about complexity. I checked the suggested duplications, but most of them are actually wrong. So it's fine from my side now.

eudesbarbosa (commented):
@mkuhring , you have pairs of analysis files like `/annotate_input/a_isatest_transcriptome_profiling_nucleotide_sequencing_{1,2}.txt`. 
What is the difference between the two? Are they used in the same context?

cubi_tk/isa_tab/__init__.py
```diff
@@ -13,6 +13,57 @@
...
+By default, if Characteristics or Comments with the same name already exist for a node type, only empty values are
+updated. Overwriting existing values requires confirmation (`--force-update`).
+
+Annotations are only applied to only one study and assay, since material names are not necessarily unique between the
+same material types of different studies or different assays (and thus, annotations couldn't be assigned unambiguously).
+By default the first study and assay listed in the investigation file are considered for annotation. A specific study
+and assay may be selected by file name (not path!) via `--target-study` or `--target-assay`, resp.
+
+Example execution:
+``cubi-tk isa-tab annotate investigation.tsv annotation.tsv --target-study s_study.tsv --target-assay a_assay.tsv``
```
eudesbarbosa, line 40: I'm not sure I understand this example. Which files are local? How one would find out the `--target-study` file name?
Also, do I need to be in a specific directory to call this command?

mkuhring:
@eudesbarbosa, the analysis files are actually just duplicates, but I use them to test assay selection. E.g. to see that I in fact annotate the second and not the first assay.

cubi_tk/isa_tab/annotate.py
```diff
@@ -0,0 +1,458 @@
...
+
+        return None
+
+    def _has_content(self, value):
+        if is_ontology_term_ref(value):
+            return value.name or value.accession or value.ontology_name
+        else:
+            return value
+
+    def _update_characteristics(self, material, assay):
```
eudesbarbosa, line 78: This is complex method and definitely would benefit from more comments. Just by reading the code I cannot easily understand what are the cases in the if-else statements. particularly, the different in these returns:
```
 if assay and material.type == SAMPLE_NAME:
   return material

 ...

else:
   return material
```

Further, can this method be further divided? It seems like it dealing with different parts: `elif` block for comments, and another for header.

tests/test_isa_tab_annotate.py
```diff
@@ -0,0 +1,87 @@
...
+    # test reference files
+    path_test = os.path.join(os.path.dirname(__file__), "data", "isa_tab", "annotate_result1")
+
+    # tests
+    files = glob.glob(os.path.join(path_test, "*"))
+
+    match, mismatch, errors = filecmp.cmpfiles(
+        path_test, path_input, (os.path.basename(f) for f in files), shallow=False
+    )
+    print([match, mismatch, errors])
```
eudesbarbosa, line 44: Do we have a standard for outputting debug statements?

eudesbarbosa (commented):
@mkuhring , in general it looks good: clean code and well structured :)

It would just benefit from more documentation associated with the code. As a new guy I would have a very hard time understanding what the methods suppose to be doing and the expected variant types in the methods.

eudesbarbosa:
> 
> 
> @eudesbarbosa, the analysis files are actually just duplicates, but I use them to test assay selection. E.g. to see that I in fact annotate the second and not the first assay.

Can the file names somehow reflect that?

tests/test_isa_tab_annotate.py
```diff
@@ -0,0 +1,87 @@
...
+import os
+import glob
+import filecmp
+
+from shutil import copytree
+
+from cubi_tk.__main__ import main
+
+
+def test_run_isatab_annotate_case1(tmp_path):
```
eudesbarbosa, line 13: Are these two methods the same except for the input file?

cubi_tk/isa_tab/__init__.py
```diff
@@ -13,6 +13,57 @@
...
     file and update existing ones, e.g., for newly added parents.
 
+``annotate``
+    Add annotation to an ISA-tab file, given a tsv file.
+
+
+Annotate
+--------
+
+``cubi-tk isa-tab annotate`` updates material and files nodes in ISA-tab studies and assays with annotations provided as
```
holtgrewe, line 23: These lines look pretty long. We're using a line length of 100 in the code.

cubi_tk/isa_tab/__init__.py
```diff
@@ -13,6 +13,57 @@
...
+By default, if Characteristics or Comments with the same name already exist for a node type, only empty values are
+updated. Overwriting existing values requires confirmation (`--force-update`).
+
+Annotations are only applied to one study and assay, since material names are not necessarily unique between the
+same material types of different studies or different assays (and thus, annotations couldn't be assigned unambiguously).
+By default the first study and assay listed in the investigation file are considered for annotation. A specific study
+and assay may be selected by file name (not path!) via `--target-study` or `--target-assay`, resp.
+
+Example execution:
+``cubi-tk isa-tab annotate investigation.tsv annotation.tsv --target-study s_study.tsv --target-assay a_assay.tsv``
```
holtgrewe, line 40: Prefer to use `.. code-block:` or `::`.

cubi_tk/isa_tab/annotate.py
holtgrewe: I second this, please split each branch into its own method with its own docstring.

cubi_tk/isa_tab/annotate.py
```diff
@@ -0,0 +1,457 @@
...
+                annotation_data.append(row)
+
+        # Need at least two columns (e.g. one source/sample name plus new annotation) and one data row
+        if len(annotation_data) < 2 and len(annotation_data[0]) < 2:
+            logger.error("Annotation file needs at least two columns and two rows")
+            return 1
+        return annotation_data
+
+    def _build_annotation_map(self, annotation):
+        # change to long df
```
holtgrewe, line 336: Format this comment a bit better, e.g., use `Change to long def.` (proper lower/upper case) with an empty line afterwards.

cubi_tk/isa_tab/annotate.py
```diff
@@ -0,0 +1,457 @@
...
+            # update available characteristics
+            for c in material.characteristics:
+                if c.name in annotation:
+                    if self._has_content(c.value[0]):
+                        if self.overwrite:
+                            c = attr.evolve(c, value=[annotation.pop(c.name)])
+                            # TODO: consider ontologies as values
+                        else:
+                            warn(
+                                f"Value for material {material.name} and characteristic {c.name} "
```
holtgrewe, line 96: I believe so far we are not using format strings in Python. Also this will break compatibility with older Python versions.

cubi_tk/isa_tab/annotate.py
```diff
@@ -0,0 +1,457 @@
...
+            comments = []
+            # update available comments
+            for c in material.comments:
+                if c.name in annotation:
+                    if c.value:
+                        if self.overwrite:
+                            c = attr.evolve(c, value=[annotation.pop(c.name)])
+                        else:
+                            warn(
+                                f"Value for material {material.name} and comment {c.name} "
```
holtgrewe, line 146: same

docs_manual/index.rst
```diff
@@ -12,7 +12,8 @@ Installation & Getting Started
 Manual
     This section contains manuals for specific commands.
 
-    | :ref:`Creating ISA-tab folders <man_isa_tpl>`
+    | :ref:`Creating ISA-tab files <man_isa_tpl>`
+    | :ref:`Adapting ISA-tab files <man_isa_tab>`
```
holtgrewe, line 16: Adapting, not annotating?

holtgrewe (changes requested):
Overall looks almost done. Please address my and Eudes' points.

eudesbarbosa:
@mkuhring , way better. Thanks for the update :)

PR #55, opened 2021-07-01
Minor fixes

holtgrewe:


holtgrewe:
@eudesbarbosa @ericblanc20 some small fixes... By the merge conflict that I had to resolve, it becomes apparent to me that we need to harmonize the sequencing type vocabulary.

eudesbarbosa (approved):
LGTM

...maybe rebase this one instead of squash?

ericblanc20 (commented):
RNA data is a problem to me, because there are so many different assay types.

How should we harmonize cubi-tk with [biomedsheets](https://github.com/bihealth/biomedsheets/blob/306b67242ceb93baa3dd90fac91eb1189f93e807/biomedsheets/io_tsv/base.py#L38-L39)?

holtgrewe:
> How should we harmonize cubi-tk with biomedsheets?

I think we should just use whatever biomedsheets uses for now. So I need to move from `RNA_seq` to `mRNA_seq`.

PR #61, opened 2021-07-15
Enable missing data in pull-raw-data

ericblanc20:
Changes to `cubi-tk/sodar/pull_raw_data.py` to enable downloading raw data from assays for which some samples don't (yet) have data.

Notes:

- The test for data availability is done using the iRODS command `ils`. It is probably not the most reliable nor efficient way.
- When the `--allow-missing` option is set, the list of samples without data is provided as warning. However, the warning is shown just before the list of samples to download, so it could be missed by the user. I am not sure what is the best way to improve on that.
- There are still no automatic test for the new feature, I am not yet able to write proper ones.

### DIFF
--- a/cubi_tk/sodar/pull_raw_data.py
+++ b/cubi_tk/sodar/pull_raw_data.py
@@ -5,7 +5,7 @@ import os
 import shlex
 import typing
 from pathlib import Path
-from subprocess import SubprocessError, check_call
+from subprocess import SubprocessError, check_call, DEVNULL
 
 import attr
 from logzero import logger
@@ -26,6 +26,7 @@ class Config:
     sodar_api_token: str = attr.ib(repr=lambda value: "***")  # type: ignore
     overwrite: bool
     min_batch: int
+    allow_missing: bool
     dry_run: bool
     irsync_threads: int
     yes: bool
@@ -106,6 +107,10 @@ class PullRawDataCommand:
         parser.add_argument("--min-batch", default=0, type=int, help="Minimal batch number to pull")
 
         parser.add_argument(
+            "--allow-missing", default=False, action="store_true", help="Allow missing data in assay"
+        )
+
+        parser.add_argument(
             "--yes", default=False, action="store_true", help="Assume all answers are yes."
         )
         parser.add_argument(
@@ -171,6 +176,15 @@ class PullRawDataCommand:
                     break
 
         library_to_folder = self._get_library_to_folder(assay)
+
+        if self.config.allow_missing:
+            commands = self._build_check_commands(assay, library_to_folder)
+            no_data = self._executed_check_commands(commands)
+            for k in no_data:
+                logger.warning("No data for library %s, ignored...", k)
+                del(library_to_folder[k])
+
+
         commands = self._build_commands(assay, library_to_folder)
         if not commands:
             logger.info("No samples to transfer with --min-batch=%d", self.config.min_batch)
@@ -187,6 +201,12 @@ class PullRawDataCommand:
                 "i:%s/%s" % (assay.irods_path, k),
                 "%s/%s" % (self.config.output_dir, v),
             ]
+        return commands
+
+    def _build_check_commands(self, assay, library_to_folder):
+        commands = {}
+        for k in library_to_folder.keys():
+            commands[k] = ["ils", "%s/%s" % (assay.irods_path, k)]
         return commands
 
     def _executed_commands(self, commands):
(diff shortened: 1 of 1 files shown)


eudesbarbosa (changes requested):
@ericblanc20, please fix the issues from [Codacy](https://app.codacy.com/gh/bihealth/cubi-tk/pullRequest?prid=7759826) and review the code style: `black --check --line-length 100 .`

cubi_tk/sodar/pull_raw_data.py
```diff
@@ -171,6 +176,15 @@ def execute(self) -> typing.Optional[int]:
                     break
 
         library_to_folder = self._get_library_to_folder(assay)
+
+        if self.config.allow_missing:
+            commands = self._build_check_commands(assay, library_to_folder)
+            no_data = self._executed_check_commands(commands)
+            for k in no_data:
```
eudesbarbosa, line 183: Use meaningful variable name instead of `k`.

cubi_tk/sodar/pull_raw_data.py
```diff
@@ -189,6 +203,12 @@ def _build_commands(self, assay, library_to_folder):
             ]
         return commands
 
+    def _build_check_commands(self, assay, library_to_folder):
```
eudesbarbosa, line 206: Please keep code documented. Example:
```
"""Method summary.

:param assay: Assay definition.
:type assay: Assay type.

...

:returns: Explanation of what the method returns.
"""
```

Also, can't this be a `@staticmethod`?

cubi_tk/sodar/pull_raw_data.py
```diff
@@ -215,6 +235,16 @@ def _executed_commands(self, commands):
                     return 1
         return 0
 
+    def _executed_check_commands(self, commands):
```
eudesbarbosa, line 238: Same documentation issue, plus Codacy issue.

cubi_tk/sodar/pull_raw_data.py
```diff
@@ -189,6 +203,12 @@ def _build_commands(self, assay, library_to_folder):
             ]
         return commands
 
+    def _build_check_commands(self, assay, library_to_folder):
+        commands = {}
+        for k in library_to_folder.keys():
```
eudesbarbosa, line 208: Please use meaningful variables name instead of `k`.

eudesbarbosa (commented):
@ericblanc20, have you considered "asking for forgiveness instead of permission" kind of approach? Reading the full code it looks like you could simply use your new `allow_missing` in the exception part of `_executed_commands()`:

```

    def _executed_commands(self, commands):
        cmds_txt = "\n".join(["- %s" % " ".join(map(shlex.quote, cmd)) for cmd in commands])
        ...
                else:
            for cmd in commands:
                try:
                    cmd_str = " ".join(map(shlex.quote, cmd))
                    ...
                    check_call(cmd)
                except SubprocessError as e:  # pragma: nocover
                    if not self.config.allow_missing:
                         logger.error("Problem executing irsync: %s", e)
                         return 1
                    else:
                         <check if correct error>:
                               <warning message>
        return 0
```

ericblanc20:
- I followed Eudes advice (thanks!), and catch the `irsync` error, rather than checking folder existence with `ils`.
- Now the `_executed_commands` method from class `PullRawDataCommand` returns a list of failed transfers, rather than a `0/1` status. Is that acceptable?
- FYI, `black --check --line-length 100` doesn't report any error on file `cubi-tk/sodar/pull_raw_data.py` for `black` version `21.5b2`.

cubi_tk/sodar/pull_raw_data.py
```diff
@@ -171,26 +179,38 @@ def execute(self) -> typing.Optional[int]:
                     break
 
         library_to_folder = self._get_library_to_folder(assay)
+
         commands = self._build_commands(assay, library_to_folder)
         if not commands:
             logger.info("No samples to transfer with --min-batch=%d", self.config.min_batch)
             return 0
-        return self._executed_commands(commands)
```
eudesbarbosa, line 178: This seems out of place. I'd leave all this logic inside `_executed_commands()`, the rest of the code just need to know if fails or not.

cubi_tk/sodar/pull_raw_data.py
```diff
@@ -200,20 +220,24 @@ def _executed_commands(self, commands):
...
                     print(cmd_str)
                     check_call(cmd)
                 except SubprocessError as e:  # pragma: nocover
-                    logger.error("Problem executing irsync: %s", e)
-                    return 1
-        return 0
+                    if not self.config.allow_missing:
+                        logger.error("Problem executing irsync: %s", e)
+                        return [lib]
+                    else:
```
eudesbarbosa, line 238: I only thought about it after I give you the suggestion, but to keep it robust you need to include the exact error code from iRODS (there are many reasons the transfer would fail).
```

   if self.config.allow_missing and e.returncode = <iRODS error code you are expecting>:
        <warning messages logic>
   else:
          logger.error("Problem executing irsync: %s", e)
          return 1
   ...
```

eudesbarbosa (changes requested):
Looking good, Eric. I included a few comments to further simply it.

Regarding `black`, there is still something wrong with it. You can look into the 'Checks' tab to see where the automatic tests failed.

eudesbarbosa (approved):
Looks good, thanks for the changes!

Issue #59, opened 2021-07-07
Refurbish cubi-tk irods check


holtgrewe:
**Is your feature request related to a problem? Please describe.**
The current implementation is not so practical (any more). For example, it actually downloads the files and computes md5 check sums which is not practical for really large directories and uses icommands in too many places rather than the Python API.

**Describe the solution you'd like**
Change it as follows:

- Use Python API for the parts that don't do heavy lifting
- Perform checks:
    - count replicas
    - check that md5 sum exists for all replicas and is consistent
    - check that the file length of all replicas is the same
    - check that 0-sized files have the md5sum of the empty string `d41d8cd98f00b204e9800998ecf8427e`
- Optionally, download file and compute md5 sums (opt-in)

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

mikkonie:
We don't actually have to separately check the file size, since we're already comparing checksums.

PR #65, opened 2021-08-17
Reimplement irods check using python client (#59)

mikkonie:
Here's my solution for re-implementing the iRODS check using python-irodsclient instead of shell hacks.

Some notes:

- For opening a session via the client we require the user's iRODS password. My solution was to prompt for this in CLI as storing passwords as clear text is a no-no. I'm open for better ideas.
- Based on my own testing (which includes running this on an iRODS collection with 2500*2 files) opening several sessions and running in parallel does not seem to speed up the checks much at all. We might want to ditch the parallel stuff altogether to simplify the code, but that's your call.
- The alternate version of irods check under sea-snap does not work with the initial commit. If you give thumbs up to this one I should be quite easily able to refactor the hard coded parts to work similarly.
- Codacy doesn't like us providing raw SQL to `SpecificQuery()`. That's too bad, but not much I can do. It's the only efficient way to get file listings recursively. `walk()` is horribly inefficient and becomes unusable beyond a couple of subcollections. Been there, tried that.
- If I have accidentally broken some cubi-tk conventions, please let me know :P

### DIFF
--- a/cubi_tk/irods/check.py
+++ b/cubi_tk/irods/check.py
@@ -1,19 +1,14 @@
 """``cubi-tk irods check``: Check target iRODS collection (all md5 files? metadata md5 consistent? enough replicas?)."""
 
 import argparse
-import getpass
 import json
 import os
-import random
 import re
-import string
-import sys
 import tqdm
 import typing
 
-from irods.exception import CAT_NO_ROWS_FOUND
-from irods.models import Collection, DataObject
-from irods.query import SpecificQuery
+from irods.collection import iRODSCollection
+from irods.data_object import iRODSDataObject
 from irods.session import iRODSSession
 from logzero import logger
 from multiprocessing.pool import ThreadPool
@@ -32,18 +27,19 @@ class IrodsCheckCommand:
         #: Command line arguments.
         self.args = args
 
+        #: Path to iRODS environment file
+        self.irods_env_path = os.path.join(
+            os.path.expanduser("~"), ".irods", "irods_environment.json"
+        )
+
         #: iRODS sessions for parallel execution
         self.irods_sessions = []
 
         #: iRODS environment
         self.irods_env = None
 
-        #: iRODS password
-        self.irods_pass = None
-
     def __del__(self):
         # Ensure cleanup of iRODS sessions
-        # NOTE: Possibly not necessary anymore with python-irodsclient v1.0.0?
         for irods in self.irods_sessions:
             irods.cleanup()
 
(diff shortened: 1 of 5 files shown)


mikkonie:
I made some updates. Details below:

- Upgraded python-irodsclient to get past the issues with expired certs.
- Interactive password is no longer required, we now use the stored iRODS password.
- We now use `walk()` instead of `SpecificQuery`.
    * Apparently `SpecificQuery` does not work with a `rodslevel` user, as we always get `SYS_NO_API_PRIV` from iRODS. This makes sense, as this could enable a standard user to break a lot of things via direct database queries.
    * `walk()` is HORRIBLY inefficient, which means this approach is much slower than my initial approach if we have more than a few files/subcollections. Don't really know how to help it.
- However, in this experience multiple threads do help quite a bit, so I left them in.
- I guess I still have to fix up the snappy version.
- Black formatting fails in CI because of the dev install of python-irodsclient. `--exclude "src"` might work..

cubi_tk/irods/check.py
```diff
@@ -1,73 +1,100 @@
...
+            os.path.expanduser("~"), ".irods", "irods_environment.json"
+        )
+
+        #: iRODS sessions for parallel execution
+        self.irods_sessions = []
+
+        #: iRODS environment
+        self.irods_env = None
+
+    def __del__(self):
```
holtgrewe, line 41: This looks like it's prone to concurrency/inderministic issues. Can we switch this to use Python context managers that ensure that the sessions are closed / cleaned more deterministically?

cubi_tk/irods/check.py
```diff
@@ -1,73 +1,100 @@
...
-                    base_path = g[2][:-1]
-                elif not g[1]:
-                    ftype = "files" if g[2][-4:] != ".md5" else "md5"
-                    files[ftype].append(os.path.join(base_path, g[2]))
-        return files
+
+    def get_data_objs(self, root_coll: iRODSCollection):
+        """Get data objects recursively under the given iRODS path."""
+        # NOTE: We shouldn't use walk() as it's extremely inefficient, but
+        #       calling SpecificQuery with a rodsuser results in SYS_NO_API_PRIV
```
holtgrewe, line 89: Can we ask on the irods support/mailing list/github issue about the `SYS_NO_API_PRIV` issue?

cubi_tk/irods/check.py
```diff
@@ -1,73 +1,100 @@
...
+        data_objs = dict(files=[], md5s={})
+        for res in root_coll.walk():
+            for obj in res[2]:
+                if obj.path.endswith(".md5"):
```
holtgrewe, line 93: Can we make the `.md5` part configurable and allow the user to specify `sha256` as an argument?

cubi_tk/irods/check.py
```diff
@@ -84,87 +111,104 @@ def execute(self):
...
-        logger.info("Checking %s files (first 20 shown):\n%s", num_files, lst_files)
+    def run_checks(self, irods: iRODSSession, data_objs: dict):
+        """Run checks on files, in parallel if enabled."""
+        num_files = len(data_objs["files"])
+        lst_files = "\n".join([f.path for f in data_objs["files"]][:19])
+        logger.info(
+            "Checking %s file%s%s:\n%s",
+            num_files,
+            "s" if num_files != 1 else "",
+            " (first 20 shown)" if num_files > 20 else "",
```
holtgrewe, line 148: Can we make the `20` configurable and have `0` mean "no limit"?

cubi_tk/irods/check.py
```diff
@@ -84,87 +111,104 @@ def execute(self):
...
-            "File checksum not consistent with md5 file...\n"
-            "file: %s\n.md5-file checksum: %s\n"
-            "metadata checksum: %s",
-            file,
-            md5sum,
-            meta_info[0]["data_checksum"],
+
+    # 2) Checksums of all replicas consistent with .md5 file?
+    else:
+        with md5_obj.open("r") as f:
```
holtgrewe, line 191: Can we use open mode `"rt"` to avoid the decoe below?

cubi_tk/sea_snap/check_irods.py
```diff
@@ -77,7 +77,7 @@ def execute(self):
 
         # --- get lists
         # files on SODAR
-        files = self.get_files()
+        files = self.get_data_objs()
         files_rel = [f.replace(self.args.irods_path + "/", "") for f in files["files"]]
         logger.info("Files on SODAR (first 20): %s", ", ".join(files_rel[:19]))
```
holtgrewe, line 82: Again, can we make 20 configurable?

requirements/base.txt
```diff
@@ -63,3 +63,7 @@ simplejson
 
 # REST API client for SODAR
 sodar-cli ==0.1.0
+
+# Python iRODS client
+# python-irodsclient>=1.0.0, <1.1
+-e git+https://github.com/irods/python-irodsclient.git@e00523bcc159d4e55a7d1920a0bb8e345bf38796#egg=python-irodsclient
```
holtgrewe, line 69: No release yet? :-)

holtgrewe (approved):
Looks good, see my comments below. I see the context manager vs __del__ one as the most important one.

cubi_tk/irods/check.py
mikkonie: I realized after pushing this commit that it's not exactly an issue: SpecificQuery basically allows writing any SQL to the iRODS database. I think there's a good reason why regular users are not allowed to do this. Wouldn't want a user to drop tables accidentally or maliciously. So walk() is, sadly, the way to go.

cubi_tk/irods/check.py
mikkonie: I guess if I refactor the whole thing into such that we open new connections for each operation with a context manager and then do a cleanup, that could work. It will probably be slower though. Not sure what the issue here is.

cubi_tk/irods/check.py
mikkonie: Technically yes, it's not configurable in SODAR though.

cubi_tk/irods/check.py
mikkonie: Why not. That part wasn't done by me to begin with :)

cubi_tk/irods/check.py
mikkonie: Maybe? Can try.

cubi_tk/sea_snap/check_irods.py
mikkonie: Yes, will do.

requirements/base.txt
mikkonie: Nope, still not.

cubi_tk/irods/check.py
holtgrewe: Something like this could wokr

```
from contextlib import contextmanager

class IrodsCheckCommand:

    # ...

    @contextmanager
    def _irods_sessions(self, first, count):
        irods_sessions = [first]
        irods_sessions += [self.connect() for _ in range(count - 1)]
        try:
            yield irods_sessions
        finally:
            for irods in irods_sessions:
                irods.cleanup()


    # ...

        # ...
        with self._irods_sessions(session, s_count) as irods_sessions:
            pass  # ...
```

mikkonie: Alright, will refactor.

mikkonie: Done.

cubi_tk/irods/check.py
mikkonie: Done.

cubi_tk/irods/check.py
mikkonie: Done.

cubi_tk/sea_snap/check_irods.py
mikkonie: Oh, this is in the sea-snap version. Haven't touched it, not sure how it should be modified or how to test it since I've never dealt with sea-snap. Hints welcome :)

PR #66, opened 2021-09-22
modified the isa-tpl/generic to make SODAR plugin work

january3:
modified the isa-tpl/generic such that the SODAR plugin for Library N…ame works correctly out of the box

The problem was in the lacking sources / accessions of the study assay measurement.


eudesbarbosa (approved):
LGTM!

PR #69, opened 2021-10-05
Add assay support in itransfer_common

ericblanc20:
- `--assay` option in `itransfer_common.py`, so used by all `itransfer` subcommands from the `snappy` command.
- make `black` happy with docstrings (`cubi_tk/sea_snap/write_sample_info.py`)

### DIFF
--- a/cubi_tk/snappy/itransfer_common.py
+++ b/cubi_tk/snappy/itransfer_common.py
@@ -183,6 +183,9 @@ class SnappyItransferCommandBase:
             help="After files are transferred to SODAR, it will proceed with validation and move.",
         )
         parser.add_argument(
+            "--assay", dest="assay", default=None, help="UUID of assay to download data for."
+        )
+        parser.add_argument(
             "destination", help="UUID from Landing Zone or Project - where files will be moved to."
         )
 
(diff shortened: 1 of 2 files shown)


cubi_tk/snappy/itransfer_common.py
```diff
@@ -380,15 +383,15 @@ def get_sodar_info(self):
                 # Behavior: search for available LZ; if none,create new LZ.
                 try:
                     lz_uuid, lz_irods_path = self.get_latest_landing_zone(
-                        project_uuid=in_destination
+                        project_uuid=in_destination, assay_uuid=self.args.assay
```
eudesbarbosa, line 386: Just to keep it consistent with the rest of the code I would suggest a copy of the variable in `# Initialise variables` - Just to make clear which variables are important for the method right from the jump.
Like: 
```
        not_project_uuid = False
        create_lz_bool = self.args.yes
        assay_uuid = self.args.assay  # <<< HERE
        in_destination = self.args.destination

```

eudesbarbosa (changes requested):
Looks good, I just made a small request regarding style. 

Also, and this part is completely optional, if `SampleInfoTool` is your class and you feel like, you could further improve the methods documentation with more info about the arguments - same deal as you did for `create_landing_zone()`.

eudesbarbosa (approved):
LGTM

PR #72, opened 2021-11-01
Fix typo in usecase_exomes

messersc:


eudesbarbosa (approved):
@messersc, thanks for improving the documentation. 

**Note:**  Merging before required statuses passed as the issue was solved in #73

PR #75, opened 2021-11-17
Add generic itransfer-step

ericblanc20:
**Aim:** Generic function to add the output of any `snappy` step to SODAR.

**Remarks:**

- In the somatic/cancer branch, many steps should be added to SODAR: `somatic_variant_calling`, `somatic_variant_annotation`, `somatic_targeted_seq_cnv_calling`, `somatic_wgs_cnv_calling`, perhaps `hla_typing`, `gene_expression_quantification`, `somatic_gene_fusion_calling`, and others that will be implemented "soon".
- Supersedes `itransfer-ngs-mapping` & `itransfer-variant-calling`.
- Implementation is not ideal: missing step names are not caught in `cubi_tk.snappy.itransfer_common.check_args`, but in `cubi_tk.snappy.itransfer_common.execute`. This is because `cubi_tk.sodar.ingest_fastq` creates a sub-class of `SnappyItransferCommandBase`, without defining `step_name` and without creating a `check_args` method for the subclass. I don't think that it should, but it creates the problem described above.
- The test is copied-pasted from `test_snappy_itransfer_ngs_mapping`.

### DIFF
--- a/cubi_tk/snappy/itransfer_step.py
+++ b/cubi_tk/snappy/itransfer_step.py
@@ -1,0 +1,42 @@
+"""``cubi-tk snappy itransfer-step``: transfer step results into iRODS landing zone."""
+
+import argparse
+import os
+import typing
+
+from logzero import logger
+
+from .itransfer_common import SnappyItransferCommandBase
+
+
+class SnappyItransferStepCommand(SnappyItransferCommandBase):
+    """Implementation of snappy itransfer command for results from any step."""
+
+    fix_md5_files = True
+    command_name = "itransfer-step"
+    step_name = None
+
+    @classmethod
+    def setup_argparse(cls, parser: argparse.ArgumentParser) -> None:
+        super().setup_argparse(parser)
+        parser.add_argument(
+            "--step", help="Name of the snappy pipeline step", default="ngs_mapping"
+        )
+        parser.add_argument(
+            "--tool", help="Name of the tool, for example bwa. Tools order in important", nargs="*"
+        )
+
+    def build_base_dir_glob_pattern(self, library_name: str) -> typing.Tuple[str, str]:
+        prefix = ".".join(self.args.tool)
+        logger.debug("Using prefix {}".format(prefix))
+        return (
+            os.path.join(
+                self.args.base_path, self.step_name, "output", prefix + "." + library_name
+            ),
+            "**",
+        )
+
+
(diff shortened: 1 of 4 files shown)


cubi_tk/snappy/itransfer_step.py
```diff
@@ -0,0 +1,42 @@
...
+
+    fix_md5_files = True
+    command_name = "itransfer-step"
+    step_name = None
+
+    @classmethod
+    def setup_argparse(cls, parser: argparse.ArgumentParser) -> None:
+        super().setup_argparse(parser)
+        parser.add_argument(
+            "--step", help="Name of the snappy pipeline step", default="ngs_mapping"
```
eudesbarbosa, line 23: Shouldn't the default be `None`? I'd argue that it is better to fail than having users accidentality transferring ngs data.
You can probably also improve the help by giving examples. I never found obvious the relation between name of step and directory.

tests/test_snappy_itransfer_step.py
```diff
@@ -0,0 +1,156 @@
...
+        parser.parse_args(["snappy", "itransfer-step", "--help"])
+
+    assert e.value.code == 0
+
+    res = capsys.readouterr()
+    assert res.out
+    assert not res.err
+
+
+def test_run_snappy_itransfer_ngs_mapping_nothing(capsys):
```
eudesbarbosa, line 29: If you agree to the `None` suggestion for the default, perhaps you can include a new test to assert that the exit code is as expected.

tests/test_snappy_itransfer_step.py
```diff
@@ -0,0 +1,156 @@
...
+    mocker.patch("cubi_tk.snappy.itransfer_common.check_output", mock_check_output)
+
+    fake_open = fake_filesystem.FakeFileOpen(fs)
+    mocker.patch("cubi_tk.snappy.itransfer_common.open", fake_open)
+    mocker.patch("cubi_tk.snappy.common.open", fake_open)
+
+    mock_check_call = mock.mock_open()
+    mocker.patch("cubi_tk.snappy.itransfer_common.check_call", mock_check_call)
+
+    # # requests mock
```
eudesbarbosa, line 119: Please remove commented block.

eudesbarbosa (commented):
Looks good, I just included some small comments.

eudesbarbosa:
Looks good, thank you for the changes.

PR #77, opened 2021-12-07
WIP: Edit archive docs.

messersc:
Hi @ericblanc20,

I started working on the docs for cubi-tk archive. Let's track progress here until we are ready to merge.

1) What is currently a bit unclear is how to run hashdeep correctly. However, I think we should move it into ```archive prepare``` and couple it a bit tighter to the workflow, e.g. with pre-defined files and paths for the 2 hashdeep outputs.

2) What is missing right now is how to copy/move the prepared archive away from /fast. If I am correct, the steps should be
    * archive prepare
    * hashdeep - integrate if we can so there is one step less
    * archive copy
    * rsync files to ```/data/ceph-1```
    * remove archive
    * remove project

Is that correct?

Clemens


ericblanc20:
Hi @messersc,

Corrections: the steps are:

1. archive prepare
2. hashdeep of the original project - integrate if we can so there is one step less
3. archive copy
4. hashdeep audit of the copy based on the output of 2. - can be integrated if 2. is integrated
5. remove temporary destination defined in 1.
6. remove project

Integration of 2. & 4. should be easy. I will have a look at it, this afternoon hopefully, and add to the branch.

ericblanc20:
At the moment, `copy` does hashdeep on the temporary destination, and performs the audit between the temporary and final destinations.

This is not very helpful. There should be hasdeep of the original project, which should be audited against the final copy. 

In case of problems, then we might want to audit the temporary destination against the original, and/or audit the final copy against the temporary destination. That could be added as options.

cubi_tk/archive/copy.py
```diff
@@ -198,11 +258,14 @@ def _restore_relative_symlinks(self, root, rel_symlinks, add_dangling=True):
...
-                os.path.join(root, os.path.dirname(relative), destination)
-            ):
-                os.symlink(destination, os.path.join(root, relative))
+        for (relative, target) in rel_symlinks:
+            symlink_path = os.path.join(root, relative)
+            if os.path.exists(symlink_path) or os.path.islink(symlink_path):
+                continue
+            symlink_dir = os.path.dirname(symlink_path)
+            if add_dangling or os.path.exists(os.path.join(symlink_dir, target)):
+                os.makedirs(symlink_dir, mode=488, exist_ok=True)
```
messersc, line 267: mode 488 could have a comment explaining that it means permissions 750 in octal notation.

messersc (commented):
Minor comments.

PR #84, opened 2022-02-14
Added option --ignore-tar-errors so that read-protected directories d…

ericblanc20:
…on't make compress fail


eudesbarbosa (approved):
CI build failure ignored at the moment, known build issue not related to the changes in the code:
https://github.com/Thibauth/python-pushover/issues/42

PR #87, opened 2022-02-18
Fix CI dependency installation error

eudesbarbosa:
**Changes**
* Pinned `setuptools` version - breaking change in [58.0.0 ](https://setuptools.pypa.io/en/latest/history.html#v58-0-0)affecting dependencies.


eudesbarbosa:
Problem solved with dependency new release.

Issue #96, opened 2022-06-10
Move cookiecutter templates to their own repository


holtgrewe:
**Is your feature request related to a problem? Please describe.**
Having the ISA-tab cookiecutter templates inside of cubi-tk makes re-using them burdensome.

**Describe the solution you'd like**
Move them into a separate repository `cookiecutter-collection-cubi-isa-tab`. We should the usual convention of having one repository per cookiecutter here as we will always use them from within cubi-tk and not use them separately. Having one repository for each cookiecutter template too much overhead (for now).

cubi-tk should clone the current contents of the repository when being invoked with the templating command.

**Describe alternatives you've considered**
One could also clone the whole of cubi-tk and extract them where used but that is additional complexity.

**Additional context**
N/A

mikkonie:
Note regarding the use of these templates:  The one part of the Python code I actually use when using these templates in external software is the [_TEMPLATES](https://github.com/bihealth/cubi-tk/blob/master/cubi_tk/isa_tpl/__init__.py#L96) mapping, which informs me of the currently available templates.

This or something like it should probably be included in the potential separate repository. Otherwise, any software using these templates would need to create their own boilerplate code for recognizing and describing available templates.

eudesbarbosa:
Suggested first step: reorder `cubi_tk/isa_tpl/archive` - separate cookiecutter from config files.

PR #98, opened 2022-06-14
Moved archive cookiecutter templates

eudesbarbosa:
addresses #96 

**Changes**
* Moved `archive` cookiecutter directory - isolated ISA templates.
* Moved config files away from template directory.


eudesbarbosa:
@messersc, do you mind checking if I missed something?
...and also to review when we agree it is done?

eudesbarbosa:
Agreed. Added the suggestion in today's commit.

cubi_tk/archive/common.py
```diff
@@ -117,6 +118,19 @@ def traverse_project_files(directory, followlinks=True):
             yield get_file_attributes(os.path.join(path, filename), root)
 
 
+def load_variables(template_dir):
```
eudesbarbosa, line 121: I simplified this function from the `isa_tpl` version. Just to check, is the `extra` argument used for anything in the `archive`?

ericblanc20: Not in use, I don't see any situation which would require programmatically updating the configuration.

ericblanc20 (approved):

Issue #100, opened 2022-06-21
Parsing errors in ISA-Tab templates


mikkonie:
**Describe the bug**
The altamISA parser refuses to parse the following ISA-Tab templates:

- microarray
- tumor_normal_dna
- tumor_normal_triplets

**To Reproduce**
Try to load an ISA-Tab created from one of the aforementioned templates with altamISA with the default values. It is possible you may have altamISA set to not allow critical warnings, I am not sure about that.

**Expected behavior**
The ISA-Tab files should get correctly parsed.

**Additional context**
Error in the `microarray` template:
```
ISA-Tab import failed: altamISA exception in study "Investigation Title": Missing protocol reference in column 5 of file s_Template_Test.txt
```
Error in the `tumor_normal_dna` template:
```
ISA-Tab import failed: altamISA exception in study "Investigation Title": Missing protocol reference in column 38 of file s_Template_Test.txt
```
Error in the `tumor_normal_triplets` template:
```
ISA-Tab import failed: altamISA exception in study "Investigation Title": Missing protocol reference in column 38 of file s_Template_Test.txt
```

mikkonie:
All should be fixed now, closing.

PR #101, opened 2022-06-21
Fixing microarray isa-tab template (#100)

holtgrewe:


mikkonie:
After the fix, I get a similar error from an assay file:
```
Missing protocol reference in column 2 of file a_template_test_transcription_profiling_microarray.txt
```

Can have a look myself, should be trivial to fix.

mikkonie (approved):
I made a few changes, verified to work in SODAR with altamISA now. Will merge.

PR #109, opened 2022-07-01
extended single-cell-rnaseq template for variable column names

bobermayer:
allow specification of extra column names in isa-tab template


bobermayer:
ok from my side

mmilek (changes requested):
Can we add another column "SODAR status" whose values could be either "processed"/"unprocessed" or "existing"/"incoming", so that we could pool into cellranger more easily?
Also, could Sequencing Run be decapitalized to Sequencing run...?

PR #112, opened 2022-07-25
`sodar download-sheet`: deal with files not in SODAR archive root

### LINKED ISSUE #111: Download-sheet fails if investigation path returned by SODAR is not archive root
**Describe the bug**

SODAR maintains the original directory structure of imported sample sheets, as some people expect archived ISA-Tab sheets to be contained within a directory inside the zip while others may not. 

Similarly, if `i_dir_name` is set in the cubi-tk templates, SODAR preserves this in its database and returns the investigation path accordingly.

It seems that if the investigation path is returned by SODAR and  `download-sheet` creates a new archive from the JSON data, it results in investigation, study and assay files being inserted in different directories. This creates a crash.

I'm adding a dump of this crash in issue comments.

**To Reproduce**

1. Create or import a sample sheet in SODAR which places ISA-Tab files within a directory in an archive.
2. Run `download-sheet`

*Disclaimer:* I've only had time to look into this superficially, so I'm not 100% percent this is all that's needed for reproduction. Feel free to comment if you feel something is missing.

**Expected behavior**

The investigation, study and assay files should all be placed in the same directory within the created archive, or in the archive root.

**Additional context**

It is also possible that the data returned by SODAR describing the path is somehow misleading and I'm up for suggestions for future improvements. However, this part of the SODAR API has not been touched in a long time, so for now I'm considering it to be working as expected.

eudesbarbosa:
fixes #111

### DIFF
--- a/cubi_tk/sodar/download_sheet.py
+++ b/cubi_tk/sodar/download_sheet.py
@@ -146,6 +146,21 @@ class DownloadSheetCommand:
         return 0
 
     def _write_file(self, out_path, file_name, text):
+        """Write file.
+
+        :param out_path: Path to output directory.
+        :type out_path: str
+
+        :param file_name: File name as provided in SODAR. If input includes more than file name, extra information
+        is removed. File should be created in output path root.
+        Example: 'PROJECT/i_Investigation.txt' -> 'i_Investigation.txt'.
+        :type file_name: str
+
+        :param text: Text to be written in file.
+        :type text: str
+        """
+        # Remove extra info - use basename only
+        file_name = file_name.split('/')[-1]
         out_path = out_path / file_name
         if out_path.exists() and not self.config.overwrite and not self.config.dry_run:
             raise OverwriteRefusedException(


cubi_tk/sodar/download_sheet.py
```diff
@@ -146,6 +146,21 @@ def execute(self) -> typing.Optional[int]:
...
+        :param file_name: File name as provided in SODAR. If input includes more than file name, extra information
+        is removed. File should be created in output path root.
+        Example: 'PROJECT/i_Investigation.txt' -> 'i_Investigation.txt'.
+        :type file_name: str
+
+        :param text: Text to be written in file.
+        :type text: str
+        """
+        # Remove extra info - use basename only
+        file_name = file_name.split('/')[-1]
```
messersc, line 163: Is this always looking like a (posix?) path?

If yes, that's ok, otherwise we should probably use pathlib.

eudesbarbosa: So far I've seen only two options: `PROJECT/i_Investigation.txt` and `i_Investigation.txt`.
From the brief that I talk to Mikko and the little change this part went over the years, I don't expect much change.

Issue #123, opened 2022-09-15
Extend kickoff for roh_calling


holtgrewe:
**Is your feature request related to a problem? Please describe.**
The current `cubi-tk snappy kickoff` is missing `roh_calling`.

**Describe the solution you'd like**
Add more steps.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #124, opened 2022-09-15
Add roh-calling to cubi-tk snappy kickoff (#123)

holtgrewe:


eudesbarbosa (approved):
I wasn't aware of this part of the code. I wonder if there are other missing, like for the new workflows: repeat expansion, mei analysis...

holtgrewe:
> I wasn't aware of this part of the code. I wonder if there are other missing, like for the new workflows: repeat expansion, mei analysis...

Probably...

PR #141, opened 2022-11-22
New routine: `sodar check-remote` (#135)

### LINKED ISSUE #135: General check-remote function
**Background**
We need a simple and automatic way to compare local files (could be `sea-snap` output, but should work for all files with md5 sums) with the ones stored in SODAR. The solution should simply report on the absence of files and on potential differences in files based on checksum. It is out of the scoop of this project to: i) reupload conflicting files or ii) generate md5 files or provide other options to select files to check

**Possible solution**
Build upon the solution for `snappy check-remote`. 
1. Retrieve iRODS collection from SODAR using `class RetrieveIrodsCollection` found in `cubi-tk/cubi_tk/snappy/retrieve_irods_collection.py`
   The class generates a dictionary containing information on all Collections available in SODAR for a given project and assay. Key: file name (`str`); Value: list of iRODS objects (`List[IrodsDataObject]`).

2. Select files for comparison by parsing directory tree and search for all files with an associated `.md5` files. Given the diverse possible structures of projects it is not possible to generalize rules to which files should be pushed to SODAR. We can only derive from the transfer to SODAR, as it is mandatory to have an associated `.md5` files.
In the simplistic example below only `sea_snap_result.vcf` would be tested against the SODAR collections:
```
sea_snap_project
├── sea_snap_intermediate_result.vcf
├── sea_snap_result.vcf
└── sea_snap_result.vcf.md5

```

*Note:* a potential issue is that some of the symlinks don't have the same name as the original files. The ambiguity is justified by different tools requiring specific name patterns.

3. Report results could be divided in three categories: 
   - Files only found in SODAR;
   - Files only found locally;
   - Files found both locally and in SODAR: same and different checksum results.

   Additionally, it should also report duplicated files, i.e., same file transferred on multiple occasions.
   Example:
```
└── P001-N1-DNA1-WES1
    └── step_name
        ├── 2022-08-09
        │   └── out
        │     └── ...
        └── 2022-10-06
            └── out
               └── ...
```

**Additional information**
`snappy check-remote` pull request: #58
`snappy check-remote` implementation mistakes: #103, #105, #107

Nicolai-vKuegelgen:
Added a module/subfunction `check-remote` to cubi-tk sodar that allows comparison of local files (identified by accompanying .md5 file) and file in a remote SODAR project. File are considered equal if they have the same name and md5 sum.
The command outputs report classifying which files are found:
- only remotely
- only locally or
- bot remotely and locally

Optionally it is possible to:
- recalculate md5 sums of local files and compare them against stored values in .md5 files
- ignore md5 sums for file comparison and only compare by filename

Closes #135


cubi_tk/sodar/check_remote.py
```diff
@@ -0,0 +1,441 @@
...
+        :return: Returns dictionary of dictionaries:
+        key: directory path ; value: list of FileDataObject (one per file in that path)
+        """
+        logger.info("Starting raw data files search ...")
+
+        # Initialise variables
+        rawdata_structure_dict = defaultdict(list)
+
+        # Validate input
+        if not self.searchpath.exists():
```
eudesbarbosa, line 65: Shouldn't this be together with the other argument validations in `SodarCheckRemoteCommand.check_args()`?

cubi_tk/sodar/check_remote.py
```diff
@@ -0,0 +1,441 @@
...
+            remote_files_dict=remote_files_dict,
+            local_files_dict=local_files_dict,
+            filenames_only=self.args.filename_only,
+            irods_basepath=assay_path,
+        ).run()
+
+        if results:
+            logger.info("All done.")
+
+        return int(not results)
```
eudesbarbosa, line 436: I don't get it. Is there any circumstance that `FileComparisonChecker` won't return True?
Maybe this would be enough?
```python
        FileComparisonChecker(
            remote_files_dict=remote_files_dict,
            local_files_dict=local_files_dict,
            filenames_only=self.args.filename_only,
            irods_basepath=assay_path,
        ).run()

        logger.info("All done.")
        return 0
```

Nicolai-vKuegelgen: Yes, that would equivalent - I can make the change

cubi_tk/sodar/check_remote.py
Nicolai-vKuegelgen: `SodarCheckRemoteCommand.check_args()` already does check that this path exists - though I think I actually wrote this part earlier.
I guess it's better to remove it here then? Or should the class check anyway so that it - if imported somewhere else - can log the error (and retrun None). Or would it be better to not prevent that error and let it 'crash'.

eudesbarbosa: Probably remove. Ii may never happen, but if someone in the future decides to use let that person deal with it.

eudesbarbosa (approved):
LGTM

PR #156, opened 2023-02-01
Added --first-batch and --last-batch options to `cubi-tk snappy pull-sheets` (#154)

### LINKED ISSUE #154: cubi-tk snappy pull-sheets needs a --min-version
At the moment, we have to remove/comment out lines when we want to process later batches only.

Add a command as exists for pulling/pushing data.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Nicolai-vKuegelgen:
Added --first-batch and --last-batch options to `cubi-tk snappy pull-sheets`. 
All samples / lines outside of specified batches will be commented out automatically.

This should fix #154

### DIFF
--- a/cubi_tk/snappy/pull_sheets.py
+++ b/cubi_tk/snappy/pull_sheets.py
@@ -22,6 +22,9 @@ from ..common import CommonConfig, overwrite_helper, load_toml_config
 from ..isa_support import InvestigationTraversal, IsaNodeVisitor, isa_dict_to_isa_data, first_value
 from .models import load_datasets
 from .common import find_snappy_root_dir
+from .parse_sample_sheet import ParseSampleSheet
+from biomedsheets import io_tsv
+from biomedsheets.naming import NAMING_ONLY_SECONDARY_ID
 
 #: Template for the to-be-generated file.
 HEADER_TPL = (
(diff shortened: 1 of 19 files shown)


cubi_tk/snappy/pull_sheets.py
```diff
@@ -234,7 +247,12 @@ def on_visit_process(self, process, node_path, study=None, assay=None):
             )
 
 
-def build_sheet(config: PullSheetsConfig, project_uuid: typing.Union[str, UUID]) -> str:
+def build_sheet(
+    config: PullSheetsConfig,
+    project_uuid: typing.Union[str, UUID],
+    min_batch: int,
```
holtgrewe, line 253: Make consistent with arg names first batch / last batch.

holtgrewe:
This does not handle the case of parents being sequenced after the index, does it?

cubi_tk/snappy/pull_sheets.py
Nicolai-vKuegelgen: done

Nicolai-vKuegelgen:
> This does not handle the case of parents being sequenced after the index, does it?

Probably not - I haven't had cases like that yet so I'm not how they should be handled. In cases of new data the batch number of old entries is usually updated to the new batch and doing the same in a case like this would probably solve this issue as well? But I don't know if that would be done or not.

holtgrewe:
This has logic to ensure that families are complete, I believe.

https://github.com/bihealth/cubi-tk/blob/371a8d3260a537cdada7b0a70b124cda52c031d8/cubi_tk/snappy/parse_sample_sheet.py

PR #161, opened 2023-03-21
fix: archive copy hangs

### LINKED ISSUE #160: archive copy hangs
**Describe the bug**
`cubi-tk archive copy` hangs for large archived directories

**Expected behavior**
The `hashdeep` audit should complete, regardless of its final status.

**Additional context**
The cause is probably found in the way `hashdeep` is called from the python code. When there are large number of files & directories that cannot be verified (for example because of restricted permissions), the warning messages swamp `stderr`, causing the pipe to hang.
If the pipe is removed from the `subprocess.Popen` call, and `wait` is used rather than `communicate`, the problem should disappear.

ericblanc20:
The copying of the temporary destination to the final destination hangs (closes #160 ).

When many files are not copied over (for example many dangling links, or files with restriction access), then `hashdeep` outputs messages for these files on `stderr`. The old code buffers both `stdout` & `stderr`, and when the `stderr` buffer is full, the program hangs.

The new code avoid this race condition, by:

1. `stderr` is not piped into a buffer anymore in the call to `subprocess.Popen`
2. `wait` replaces `communicate`, meaning that the output is not available to the caller anymore. That was not used anyway.

This bug has also shown another weakness in the code: when the program hangs and is interrupted by the user, the local symlinks set aside during the preparation of the copy are not re-instated in the temporary destination. To avoid loosing these local symlinks, the restoring function has been added to the exit handlers.

The tests have been updated to accommodate the new treatment of symlinks.

### DIFF
--- a/tests/test_archive_prepare.py
+++ b/tests/test_archive_prepare.py
@@ -14,6 +14,7 @@ import pytest
 
 from cubi_tk.__main__ import main, setup_argparse
 from cubi_tk.common import execute_shell_commands
+
 from .test_archive_copy import sort_hashdeep_title_and_body
 
 SNAKEMAKE = re.compile("^.*\\.snakemake\\.tar\\.gz$")


mbenary (approved):
Looks good to me.

Issue #40, opened 2021-03-22
Simplify `cubi-tk isa-tpl`


eudesbarbosa:
**Bug description**

Currently `cubi-tk isa-tpl` only works if output directory is a subdirectory.
In `cubi_tk/isa_tpl/__init__.py`:
```
    output_dir = os.path.realpath(args.output_dir)
    output_base = os.path.dirname(args.output_dir)
    if os.path.exists(output_dir):  # pragma: no cover
        logger.error("Output directory %s already exists. Refusing to overwrite!", output_dir)
        return 1
    if not os.path.exists(output_base):  # pragma: no cover
        logger.error("Output path to output directory does not exist: %s", output_base)
        return 1
```

**To Reproduce**
Steps to reproduce the behaviour:
```
mkdir test_isa
cd test_isa
cubi-tk isa-tpl generic isa_tab
# Error:
# [E 210322 10:41:01] Output path to output directory does not exist:
# [E 210322 10:41:01] Something did not work out correctly.
#
# This on the other hand works:
# cubi-tk isa-tpl generic ./isa_tab
```

**Expected behaviour**
Simplify call by removing `output_base` check. This should be enough:
`cubi-tk isa-tpl generic isa_tab`

sellth:
Cannot reproduce this bug. Resolved?

PR #163, opened 2023-04-18
fix: handling of <output_dir> input argument

### LINKED ISSUE #162: improve `cubi-tk isa-tpl` output dir handling
Giving `output_dir` as a positional argument is mandatory for `cubi-tk isa-tpl`. Internally this overrides the `i_dir_name` value in cookiecutter.json [as described here](https://cookiecutter.readthedocs.io/en/stable/advanced/injecting_context.html) which is then subsequently used to set the project folder name and (depending on the template) for various other things. This leads to some unexpected behaviour though.

1. The value name `i_dir_name` needs to be present in all templates.
2. The user is prompted for input regarding this value by Cookiecutter, despite it already being a given input argument of the `cubi-tk` call.
3. If the user inputs a different `i_dir_name` when prompted, output is saved under this new folder name, ignoring the initial `output_dir` argument.
4. A trailing / in the `output_dir` argument leads to an additional subfolder being created.
5. Many templates define `i_dir_name` internally as `{{cookiecutter.investigation_title|replace(' ', '_')}}`. By just looking at the template files, it is not clear why this is changed to the `output_dir` name.

### What I propose:
- ~~Create the output dir outside of cookiecutter.~~ (cookiecutter always creates at least one dir)
- Set output dir name via private variable which is still required, but doesn't interfere with other template business.
- Replace all other uses of `i_dir_name` within templates by more meaningfully named variables.?

### Why is this a problem?
Specifically because I want to change `i_dir_name` to a [private variable](https://cookiecutter.readthedocs.io/en/stable/advanced/private_variables.html) to reduce user prompts.

sellth:
I didn't like that cubi-tk forced me to have `i_dir_name` as a cookiecutter.json variable which was also prompted to the user and could lead to unexpected behaviour if touched.

Changes in this pull request:
- Introduced new **private** cookiecutter variable `__output_dir` which is not prompted to the user and solely there to set the output dir name.
- Eliminated case where a trailing / in the `<output_dir>` argument would lead to an additional subfolder being created.
- Adapted test_isa_tpl.py to not expect files to be named after cubi-tk's `<output_dir>` argument.
- Removed all uses of `i_dir_name` in templates. Was only used to set `s_file_name` variable to `{{cookiecutter.investigation_title|replace(' ', '_')}}` (in fact mostly to `<output_dir>` though). `s_file_name` is now defined by `investigation_title` directly.

Fixes #162.


cubi_tk/isa_tpl/__init__.py
```diff
@@ -176,8 +176,8 @@ def run_cookiecutter(tpl, args, _parser=None, _subparser=None, no_input=False):
     logger.info(args)
 
     output_dir = os.path.realpath(args.output_dir)
-    output_base = os.path.dirname(args.output_dir)
-    extra_context["i_dir_name"] = os.path.basename(output_dir)
+    output_base = os.path.dirname(output_dir)
+    extra_context["__output_dir"] = os.path.basename(output_dir)
```
Nicolai-vKuegelgen, line 180: I think the change to `output_base` (now affected by realpath) could potentially break things if one invokes the command in wierd way with symlinks or a target in a completely different / nested destination?

cubi_tk/isa_tpl/isatab-stem_cell_core_bulk/cookiecutter.json
```diff
@@ -13,8 +13,9 @@
   "library_kit": "mRNA",
   "library_construction_meta": "Library selection,Target insert size",
   "sequencing_meta": "Platform,Instrument model,No. targeted reads,Paired-End",
-  "i_dir_name": "{{cookiecutter.study_title|replace(' ', '_')|replace('/', '-')}}",
+  "study_id": "{{cookiecutter.study_title|replace(' ', '_')|replace('/', '-')}}",
```
Nicolai-vKuegelgen, line 16: This doesn't match the scheme of the previous templates (which use s_file_name here ?)
Maybe worth to harmonize this across all template, when things are changed anyway

Nicolai-vKuegelgen (commented):
Looks good overall, but some things could maybe be improved.
Given that this changes all templates I'd ask for a second review / opinion anyway.

Also #40 might be related as well? Though the old bug cubi-tk isa-tpl (sometimes) doesn't create new directories seems to be gone, so maybe that can be closed anyway.

cubi_tk/isa_tpl/isatab-stem_cell_core_bulk/cookiecutter.json
sellth: Good point. I removed the study_id variable from this template and stem_cell_core_sc. 

We could also debate if these variables really need to be queried from the user or should be solely defined by things like "Study identifier" (i. e. changed to private variables). What's the use case for custom-named ISA files?

cubi_tk/isa_tpl/__init__.py
sellth: I don't see how. Tried several edge cases and it seems to work. Prior to this change `output_base` was not derived from a normalised path which is why trailing slashes did change its value.

ericblanc20: I think that you can construct instances where the change to `output_base` affects the destination of the output files.
For example, in the following directory structure:
```
a
├── b
│   └── dangling -> no_path
└── b2
    └── dangling -> ../../a_new/b/dangling
```
where `dangling` is a symlink which target is missing, then when the `output_dir` argument is set to `a/b2/dangling`, the old code will create the output files in the directory `a/b2/no_path`, while the new code will create them under `a/b/no_path`.

In that situation, and if the folder `a/b2/no_path` already exists, the old code throws an error, but not the new one, and the user may overlook the newly generated files.

However, this is pretty much the only situation I could imagine where the two versions of the code would lead to different results. And this situation is also fairly screwed up in my opinion, and I am not sure that the old way of generating files is much better.

ericblanc20 (commented):
Before approving the PR, we should agree on the best way to handle output paths in presence of symlinks.

sellth:
Thank you for looking into this. So the change would create an almost impossible to encounter edge-case while eliminating a regularly encountered edge case. :-)

ericblanc20 (approved):

PR #166, opened 2023-05-05
fix: add space between value and unit symbol for file size formatting

sellth:
As specified by ISO 31-0. ;-)

### DIFF
--- a/cubi_tk/common.py
+++ b/cubi_tk/common.py
@@ -195,9 +195,9 @@ def sizeof_fmt(num, suffix="B"):  # pragma: nocover
     """Source: https://stackoverflow.com/a/1094933/84349"""
     for unit in ["", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "Zi"]:
         if abs(num) < 1024.0:
-            return "%3.1f %s%s" % (num, unit, suffix)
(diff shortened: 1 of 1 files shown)


holtgrewe:
Hm, one could make the argument that number and unit form a token that is easier to process as terminal text. E.g., copy it by double-clicking.

Also cf. 

```
# ls -lh | head -n 1
total 2.2G
df -h | head -n 2
Filesystem                              Size  Used Avail Use% Mounted on
udev                                     32G     0   32G   0% /dev
```

holtgrewe:
Further evidenc, Python's tqdm prints by default `4.38s/it`

sellth:
Good point, but what if I only want to copy the number part? Also, there seems to be another discussion ongoing: https://github.com/tqdm/tqdm/issues/825

holtgrewe:
If you think it's actually an improvement then please go ahead and merge it. It's not worth the time discussing/thinking about.

holtgrewe (approved):

holtgrewe:
;-)

sellth:
This is mainly a nitpick of mine which I probably would've ignored if it weren't Friday afternoon. Let's wait a bit to see if there are more comments on this pressing issue. :-)

PR #168, opened 2023-05-08
feat: SODAR ingest command based on irods-pythonclient

sellth:
This adds a generic `sodar ingest` command which tries to mimic the base functionality of `irsync`, boosted by the power of using SODAR's API. Features include:

- no dependency on icommands; uses `iinit` session file if present, asks for password otherwise
- multiple source locations, single target collection
- recursive file matching
- creation of matched sub-folders in iRODS
- destination either as iRODS path or SODAR LZ UUID
- uses SODAR API for UUID destination & for querying LZ collections
- computes local checksums if missing
- triggers remote iRODS checksum computation
- scriptable; all user input can be skipped
- has a cool progress bar

Please try it out! Would appreciate lots of feedback.

I made this while learning Python, so please be patient. :-)

### DIFF
--- a/cubi_tk/sodar/ingest.py
+++ b/cubi_tk/sodar/ingest.py
@@ -241,6 +241,7 @@ class SodarIngest:
             logger.info("Nothing to do. Quitting.")
             sys.exit(0)
 
+        jobs = sorted(jobs, key=lambda x: x.path_src)
         itransfer = iRODSTransfer(irods_session, jobs)
         total_bytes = itransfer.total_bytes
         logger.info("Planning to transfer the following files:")
--- a/cubi_tk/sodar/ingest.py
+++ b/cubi_tk/sodar/ingest.py
@@ -47,11 +47,15 @@ class SodarIngest:
 
         # Get SODAR API info
         toml_config = load_toml_config(Config())
-
-        if not self.args.sodar_url:
-            self.args.sodar_url = toml_config.get("global", {}).get("sodar_server_url")
+        if toml_config:
+            config_url = toml_config.get("global", {}).get("sodar_server_url")
+            if self.args.sodar_url == "https://sodar.bihealth.org/" and config_url:
+                self.args.sodar_url = config_url
+            if not self.args.sodar_api_token:
+                self.args.sodar_api_token = toml_config.get("global", {}).get("sodar_api_token")
         if not self.args.sodar_api_token:
-            self.args.sodar_api_token = toml_config.get("global", {}).get("sodar_api_token")
+            logger.error("SODAR API token missing.")
+            sys.exit(1)
 
     @classmethod
     def setup_argparse(cls, parser: argparse.ArgumentParser) -> None:
@@ -91,7 +95,9 @@ class SodarIngest:
             help="Don't ask for permission.",
         )
         parser.add_argument(
-            "--collection", type=str, help="Target iRODS collection. Skips manual selection input."
+            "--collection",
+            type=str,
+            help="Target iRODS collection. Skips manual selection input.",
         )
         parser.add_argument(
             "sources", help="One or multiple files/directories to ingest.", nargs="+"
--- a/cubi_tk/sodar/ingest.py
+++ b/cubi_tk/sodar/ingest.py
@@ -14,6 +14,9 @@ from cubi_tk.irods_utils import TransferJob, get_irods_error, init_irods, iRODST
 
 from ..common import compute_md5_checksum, is_uuid, load_toml_config, sizeof_fmt
 
+# for testing
+logger.propagate = True
+
 
 @attr.s(frozen=True, auto_attribs=True)
 class Config:
@@ -233,12 +236,12 @@ class SodarIngest:
 
         for src in source_paths:
             try:
-                abspath = src.resolve()
+                abspath = src.resolve(strict=True)
             except FileNotFoundError:
                 logger.warning(f"File not found: {src.name}")
                 continue
             except RuntimeError:
-                logger.warning(f"Infinite loop: {src.name}")
+                logger.warning(f"Symlink loop: {src.name}")
                 continue
 
             if src.is_dir():
--- a/cubi_tk/sodar/ingest.py
+++ b/cubi_tk/sodar/ingest.py
@@ -7,6 +7,7 @@ import sys
 import typing
 
 import attr
+import logzero
 from logzero import logger
 from sodar_cli import api
 
@@ -16,6 +17,10 @@ from ..common import compute_md5_checksum, is_uuid, load_toml_config, sizeof_fmt
 
 # for testing
 logger.propagate = True
+
+# no-frills logger
+formatter = logzero.LogFormatter(fmt="%(message)s")
+output_logger = logzero.setup_logger(formatter=formatter)
 
 
 @attr.s(frozen=True, auto_attribs=True)
@@ -184,7 +189,7 @@ class SodarIngest:
 
             logger.info("Planning to create the following sub-collections:")
             for d in dirs:
-                print(f"{self.target_coll}/{str(d)}")
+                output_logger.info(f"{self.target_coll}/{str(d)}")
             if not self.args.yes:
                 if not input("Is this OK? [y/N] ").lower().startswith("y"):
                     logger.info("Aborting at your request.")
@@ -210,10 +215,10 @@ class SodarIngest:
         total_bytes = itransfer.total_bytes
         logger.info("Planning to transfer the following files:")
         for job in jobs:
-            print(job.path_src)
+            output_logger.info(job.path_src)
         logger.info(f"With a total size of {sizeof_fmt(total_bytes)}")
         logger.info("Into this iRODS collection:")
-        print(f"{self.lz_irods_path}/{self.target_coll}/")
+        output_logger.info(f"{self.lz_irods_path}/{self.target_coll}/")
 
         if not self.args.yes:
             if not input("Is this OK? [y/N] ").lower().startswith("y"):
--- a/cubi_tk/sodar/ingest.py
+++ b/cubi_tk/sodar/ingest.py
@@ -196,25 +196,42 @@ class SodarIngest:
                 {p["ipath"].parent for p in source_paths if not p["ipath"].parent == Path(".")}
             )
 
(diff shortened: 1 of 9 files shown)


cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+@attr.s(frozen=True, auto_attribs=True)
+class Config:
+    """Configuration for the ingest command."""
+
+    config: str = attr.field(default=None)
+    sodar_server_url: str = attr.field(default=None)
+    sodar_api_token: str = attr.field(default=None, repr=lambda value: "***")  # type: ignore
+
+
+class SodarIngest:
```
Nicolai-vKuegelgen, line 28: I think most of the functionality here should - ideally - be part of an abstract base class, similar to SnappyItransferCommandBase . Then both a general filetype agnostic Ingest could be derived from it and it could also be used for things like SodarIngestFastq (where the reliance on some snappy modules that load/assume biomedsheet like samplesheets is porbably not ideal anyway). 

I'm not sure how important that change is right now though, even if it's the best long term approach

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+            help="SODAR API token. Defaults to SODAR_API_TOKEN environment variable.",
+        )
+        parser.add_argument(
+            "-r",
+            "--recursive",
+            default=False,
+            action="store_true",
+            help="Recursively match files in subdirectories. Creates iRODS sub-collections to match directory structure.",
+        )
+        parser.add_argument(
```
Nicolai-vKuegelgen, line 95: This should probably be TRUE by default?

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+            "-c",
+            "--checksums",
+            default=False,
+            action="store_true",
+            help="Compute missing md5 checksums for source files.",
+        )
+        parser.add_argument(
```
Nicolai-vKuegelgen, line 102: This should definitely be TRUE by default

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+            "-K",
+            "--remote-checksums",
+            default=False,
+            action="store_true",
+            help="Trigger checksum computation on the iRODS side.",
+        )
+        parser.add_argument(
+            "-y",
```
Nicolai-vKuegelgen, line 110: Should probbaly be -y, --yes

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+                self.lz_irods_path = lz_info.irods_path
+                logger.info(f"Target iRods path: {self.lz_irods_path}")
+            else:
+                logger.info("Target landing zone is not ACTIVE.")
+                sys.exit(1)
+        else:
+            self.lz_irods_path = self.args.destination
+
+        # Build file list and add missing md5 files
+        source_paths = self.build_file_list()
```
Nicolai-vKuegelgen, line 170: It's bit unclear what the content of 'source_paths' will be here, and how it is updated/changed by  `process_md5_sums`

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+            coll = irods_session.collections.get(self.lz_irods_path)
+            for c in coll.subcollections:
+                collections.append(c.name)
+        except:
+            logger.error("Failed to query landing zone collections.")
+            sys.exit(1)
+        finally:
+            irods_session.cleanup()
+
+        if self.args.collection is None:
```
Nicolai-vKuegelgen, line 189: What if one wants to upload to multiple or all available collections?

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+                    bytes=p["spath"].stat().st_size,
+                )
+            )
+
+        return tuple(sorted(transfer_jobs))
+
+    def compute_irods_checksums(self, session, jobs: typing.Tuple[TransferJob, ...]):
+        for job in jobs:
+            if not job.path_src.endswith(".md5"):
+                print(
```
Nicolai-vKuegelgen, line 349: Why print ?

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+                    str(
+                        Path(job.path_dest).relative_to(f"{self.lz_irods_path}/{self.target_coll}/")
+                    )
+                )
+                try:
+                    data_object = session.data_objects.get(job.path_dest)
+                    data_object.chksum()
+                except Exception as e:
```
Nicolai-vKuegelgen, line 357: This will catch all Exceptions (i.e. network related), not just checksum related ones?

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,366 @@
...
+            logger.info("Sub-collections created.")
+
+        # Build transfer jobs
+        jobs = self.build_jobs(source_paths)
+
+        # Final go from user & transfer
+        total_bytes = sum([job.bytes for job in jobs])
+        logger.info("Planning to transfer the following files:")
+        for job in jobs:
+            print(job.path_src)
```
Nicolai-vKuegelgen, line 245: Should probably user logger not print here

Nicolai-vKuegelgen (commented):
I think this looks good overall, a few comments inside the code.
Otherwise,  I have some points that can probably be improved, though some of these are likely more of a personal wishlist than necessary changes:

- Given that this provides a new interface to irods, it might be worth to implement an abstract base class that allows easier implementation of specific ingest commands (like SodarIngestFastq)
- It looks to me like this can currently only ingest files for one collection / one sample? If so, then this will impact usability quite a bit.
- I think a general ingest function should be able to filter the files that will be uploaded  to Sodar somehow, i.e. by having the option to exclude or include only certain filetypes.
- There are several useful looking flags in the abstract `SnappyItransferCommandBase` that are not implement here, some might be worth to add:
 --num-parallel-transfer; not sure if this is specific to the way files are transferred? 
 --remote-dir-date & --remote-dir-pattern; addition of the upload date to the sub-collection path in irods/sodar is common practice and should be possible
  --validate-and-move; this & other sodar related functionalty could maybe be bundeled in one or another base class / MixIn ?

sellth:
Thanks Nicolai for going through my scribbles. Let me address some of your points right now and some more tomorrow once I'm back at the office and had time to think it through. 

Let me start with my guiding principles for this project:
- keep it simple feature wise
- make it agnostic to the special requirements of different assay types
- remove outside dependencies (i. e. icommands, md5sum)

I guess, it's the KISS principle.

>     * Given that this provides a new interface to irods, it might be worth to implement an abstract base class that allows easier implementation of specific ingest commands (like SodarIngestFastq)

Would make sense.

>     * It looks to me like this can currently only ingest files for one collection / one sample? If so, then this will impact usability quite a bit.

Only one target location, yes. Otherwise additional logic would be needed to determine which files should go where. Following the KISS principle this should probably be done by an outside script (using --collection and -y parameters) or a wrapper class. The original `rsync` also only does one destination.

> I think a general ingest function should be able to filter the files that will be uploaded to Sodar somehow, i.e. by having the option to exclude or include only certain filetypes.

File system globbing is already an option (i.e. `source/*.csv`), but I see some added value by dedicated `--exclude` and `--include` flags.

>       --num-parallel-transfer; not sure if this is specific to the way files are transferred?

irods-pythonclient uses [parallel transfer](https://github.com/irods/python-irodsclient#parallel-transfer) for PUTs of bigger files. Overall transfer speed is 40-60 MB/s, so not too shabby.

>       --remote-dir-date & --remote-dir-pattern; addition of the upload date to the sub-collection path in irods/sodar is common practice and should be possible

I would argue this should be done to the local file structure and not by the ingest command. At least not a generic one. KISS principle.

>       --validate-and-move; this & other sodar related functionalty could maybe be bundeled in one or another base class / MixIn ?

This is already implemented as `cubi-tk sodar lz_move`.

cubi_tk/sodar/ingest.py
sellth: Not sure I agree as this starts writing a bunch of files to your computer. I thought about creating temporary .md5 files and removing them during a cleanup step. Maybe that would be a better option.

cubi_tk/sodar/ingest.py
sellth: agreed

cubi_tk/sodar/ingest.py
Nicolai-vKuegelgen: I think I placed this comment unclearly. I meant the following line here (validating the checksums in irods)

sellth: I'm a bit hesitant to introduce too much "magic", in this case assuming that the user always wants to compute remote checksums right away. For large files this takes a significant amount of time. There's also an upcoming SODAR feature which would make this flag mostly irrelevant: bihealth/sodar-server/issues/1634

cubi_tk/sodar/ingest.py
sellth: I've extracted much of the iRODS logic into cubi_tk/irods_utils.py. Please have a look.

cubi_tk/sodar/ingest.py
sellth: Now always computes MD5, stores it in the TransferJob object and creates temporary files during transfer.

cubi_tk/sodar/ingest.py
sellth: This method was removed and the md5 summing moved to build_jobs.

cubi_tk/sodar/ingest.py
sellth: Again, I'm for keeping it simple and leave this kind of functionality to outside scripts or other, more specialised commands.

cubi_tk/sodar/ingest.py
sellth: Only reason: I like the way it looks.

cubi_tk/sodar/ingest.py
sellth: Again: I like the way it looks.

PR #174, opened 2023-06-01
feat: Remove hardcoded snappy-pipeline steps

### LINKED ISSUE #165: `cubi-tk snappy`is not yet adapted to new/renamed snappy steps
**Describe the bug**
The snappy step dependencies in cubi-tk/snappy/common do currently not fully match the state of snappy. This means that i.e. `cubi-tk snappy kickoff`will not start certain steps, other tools may be affected as well.

These currently defined steps have been renamed in or removed from snappy:
- roh_calling [removed without replacement?]
- variant_export
- targeted_seq_cnv_calling
- targeted_seq_cnv_annotation
- targeted_seq_cnv_export
- wgs_sv_calling
- wgs_sv_annotation
- wgs_sv_export
- wgs_cnv_calling
- wgs_cnv_annotation
- wgs_cnv_export

These steps in snappy are likely replacements or additions that may need definition in cubi-tk
- helper_gcnv_model_targeted & helper_gcnv_model_wgs
  - newly added steps (?)
- sv_calling_targeted
  - replaces: targeted_seq_cnv_calling
- sv_calling_wgs
  - replaces: wgs_sv_calling & wgs_cnv_calling
- varfish_export
  - replaces: wgs_sv_export & wgs_cnv_export & (?) targeted_seq_cnv_export
- variant_annotation
  - replaces: wgs_sv_annotation, wgs_cnv_annotation & (?) targeted_seq_cnv_annotation
- variant_export_external
  - likely a replacement for  wgs_cnv_export_external & wgs_sv_export_external, though those are still in use?
- variant_checking, variant_denovo_filtration, variant_filtration, variant_phasing
   - unclear if cubi-tk support is needed?

Support for these steps was not included in cubi-tk and is probably not needed:   
- General NGS: adapter_trimming, ngs_data_qc, ngs_mapping, ngs_sanity_checking, 
- RNAseq: gene_expression_quantification, gene_expression_report
- Various: cbioportal_export, hla_typing, tcell_crg_report, igv_session_generation, panel_of_normals, repeat_analysis
- somatic_*:
  - somatic_hla_loh_calling, somatic_msi_calling, somatic_gene_fusion_calling, somatic_neoepitope_prediction, somatic_ngs_sanity_checking, somatic_purity_ploidy_estimate, somatic_targeted_seq_cnv_calling, somatic_variant_annotation, somatic_variant_calling, somatic_variant_checking, somatic_variant_expression, somatic_variant_filtration, somatic_variant_signatures, somatic_wgs_cnv_calling, somatic_wgs_sv_calling, 

**To Reproduce**
1. Create a snappy project with of the currently not supported steps
2. run `cubi-tk snappy kickoff`
The unsupported steps will not be run.

**Expected behavior**
Cubi-tk should be updated to have support for at least all steps used in routine analysis of WGS & WES data for genetics.
A likely list of steps that should ahve support is given above.

xiamaz:
This wip code removes the need to have hardcoded pipeline steps in `cubi-tk`. Instead pipeline dependencies will be automatically parsed from the installed `snappy-pipeline`.

These changes only depend on some conventions in the `snappy-pipeline` project that should be fairly static (Snakefile needs to contain a `wf` variable), as otherwise finding the correct pipeline class is difficult.

This partially fixes #165, but can be easily extended to fully support custom pipeline stage names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,14 +16,6 @@ $ conda create -n cubi-tk python=3.10
 $ conda activate cubi-tk
 ```
 
-First install snappy-pipeline, which is required for some snappy commands:
-
-```bash
-$ git clone https://github.com/bihealth/snappy-pipeline
-$ cd snappy-pipeline
-$ pip install -e .
-```
-
 Clone CUBI-TK and install.
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -16,6 +16,14 @@ $ conda create -n cubi-tk python=3.10
 $ conda activate cubi-tk
 ```
 
+First install snappy-pipeline, which is required for some snappy commands:
+
+```bash
+$ git clone https://github.com/bihealth/snappy-pipeline
+$ cd snappy-pipeline
+$ pip install -e .
+```
+
 Clone CUBI-TK and install.
 
 ```bash
--- a/cubi_tk/snappy/common.py
+++ b/cubi_tk/snappy/common.py
@@ -2,11 +2,18 @@
 
 import pathlib
 import typing
+import pkgutil
 
 from biomedsheets import io_tsv
 from biomedsheets.naming import NAMING_ONLY_SECONDARY_ID
 from logzero import logger
 import yaml
+import importlib
+import re
+from collections import defaultdict
+
+import snappy_pipeline.workflows as snappy_workflows
+from snappy_pipeline import expand_ref
 
 #: Dependencies between the SNAPPY steps.
 DEPENDENCIES: typing.Dict[str, typing.Tuple[str, ...]] = {
@@ -36,6 +43,77 @@ class CouldNotFindPipelineRoot(Exception):
 class CouldNotFindBioMedSheet(Exception):
     """Raised when BioMedSheet could not be found in configuration file."""
 
+class DummyWorkflow:
+    """Dummy workflow that does nothing.
+    """
+    def __init__(self):
+        self.globals = defaultdict(str)
+
+    def __getattr__(self, _):
+        return self._catchall
+
+    def _catchall(self, *_, **__):
+        pass
+
+def get_available_snappy_workflows():
+    """Get all available snappy workflows in snappy_pipeline.
+
+    :return: List of ModuleInfo with individual workflow modules.
+    """
+    return list(pkgutil.iter_modules(snappy_workflows.__path__))
+
+def get_available_snappy_workflow_paths():
+    """Get module paths for all available snappy workflows.
+
+    :return: Dict of workflow name to workflow module path.
+    """
+    workflow_names = get_available_snappy_workflows()
+    return {w.name: pathlib.Path(w.module_finder.path) / w.name for w in workflow_names}
+
+def get_workflow_snakefile_object_name(snakefile_path):
+    """Find the Workflow implementation object name.
+
+    :param snakefile_path: Path to snakefile for workflow.
+    :type snakefile_path: str, pathlib.Path
+
+    :return: str Name of the implementation class or None if nothing as been found.
+    """
+
+    with snakefile_path.open() as f:
+        if m := re.search("wf\s*=\s*(\w+)\(", f.read()):
+            module_name = m.group(1)
+            return module_name
+    return None
+
+def get_workflow_step_dependencies(workflow_step_path):
+    """Find workflow dependencies for the given workflow step.
+    :param workflow_step_path: Path to the workflow step.
+    :type workflow_step_path: str, pathlib.Path
+
+    :return: List of dependencies for the given workflow step.
+    """
+    workflow_step_path = pathlib.Path(workflow_step_path)
+    step_name = workflow_step_path.name
+    step_config_path = workflow_step_path / "config.yaml"
+
+    step_module_paths = get_available_snappy_workflow_paths()
+    module_path = step_module_paths[step_name]
+    module_config_path = module_path / "Snakefile"
+    # get the name of the workflow step class name
+    obj_name = get_workflow_snakefile_object_name(module_config_path)
+    if obj_name is None:
+        raise RuntimeError(f"Could not find workflow object for {step_name}")
+
+    workflow_module = importlib.import_module("."+step_name, "snappy_pipeline.workflows")
+    workflow_class = getattr(workflow_module, obj_name)
+
+    with open(str(step_config_path), "r") as stream:
+        config_data = yaml.safe_load(stream)
+
+    config, lookup_paths, config_paths = expand_ref(str(step_config_path), config_data, [str(workflow_step_path), str(workflow_step_path.parent / ".snappy_pipeline")])
+    workflow_object = workflow_class(DummyWorkflow(), config, lookup_paths, config_paths, str(workflow_step_path))
+    dependencies = workflow_object.sub_workflows.keys()
+    return list(dependencies)
 
 def find_snappy_root_dir(
     start_path: typing.Union[str, pathlib.Path], more_markers: typing.Iterable[str] = ()
@@ -63,6 +141,19 @@ def find_snappy_root_dir(
     raise CouldNotFindPipelineRoot()
 
 
+# TODO: this assumes standard naming which is a limitation...
+# MZ: now this is easily fixable by extending the folder identification to
+# parse the config.yaml
+def get_snappy_step_directories(snappy_root_dir):
+    snappy_workflows_names = [
+        s.name for s in get_available_snappy_workflows()
+    ]
+    folder_steps = [p for p in pathlib.Path(snappy_root_dir).iterdir() if p.is_dir() and p.name in snappy_workflows_names]
+
+    return folder_steps
+
+
+
 def load_sheet_tsv(path_tsv, tsv_shortcut="germline"):
     """Load sample sheet.
 
@@ -77,6 +168,20 @@ def load_sheet_tsv(path_tsv, tsv_shortcut="germline"):
     load_tsv = getattr(io_tsv, "read_%s_tsv_sheet" % tsv_shortcut)
     with open(path_tsv, "rt") as f:
         return load_tsv(f, naming_scheme=NAMING_ONLY_SECONDARY_ID)
+
+
+def get_snappy_config(snappy_root_dir):
+    """Get snappy configuration.
+
+    :param snappy_root_dir: Path to snappy root directory.
+    :type snappy_root_dir: str, pathlib.Path
+
+    :return: Returns loaded snappy configuration.
+    """
+    snappy_config = snappy_root_dir / ".snappy_pipeline" / "config.yaml"
+    with open(snappy_config, "r") as stream:
+        config = yaml.safe_load(stream)
+    return config
 
 
 def get_biomedsheet_path(start_path, uuid):
--- a/cubi_tk/snappy/common.py
+++ b/cubi_tk/snappy/common.py
@@ -1,17 +1,10 @@
 """Common functionality for SNAPPY."""
-
-from collections import defaultdict
-import importlib
 import pathlib
-import pkgutil
-import re
 import typing
 
 from biomedsheets import io_tsv
 from biomedsheets.naming import NAMING_ONLY_SECONDARY_ID
 from logzero import logger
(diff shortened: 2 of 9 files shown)


cubi_tk/snappy/common.py
```diff
@@ -36,6 +43,77 @@ class CouldNotFindPipelineRoot(Exception):
...
+
+    step_module_paths = get_available_snappy_workflow_paths()
+    module_path = step_module_paths[step_name]
+    module_config_path = module_path / "Snakefile"
+    # get the name of the workflow step class name
+    obj_name = get_workflow_snakefile_object_name(module_config_path)
+    if obj_name is None:
+        raise RuntimeError(f"Could not find workflow object for {step_name}")
+
+    workflow_module = importlib.import_module("."+step_name, "snappy_pipeline.workflows")
```
holtgrewe, line 107: use fstring?

xiamaz:
TODO:

[ ] update docs

xiamaz:
Docs PR added in https://github.com/bihealth/snappy-pipeline/pull/400

xiamaz:
@holtgrewe This now closes #165, as support for arbitrarily named workflow step folders has been added.

Nicolai-vKuegelgen:
Meta question: is there any way to get rid of / hide these warning boxes in the code review?
image

They make the code quite unreadable :(

cubi_tk/snappy/common.py
```diff
@@ -37,6 +23,124 @@ class CouldNotFindBioMedSheet(Exception):
...
+
+    with open(str(config_path)) as stream:
+        config_data = yaml.safe_load(stream)
+
+    config, lookup_paths, config_paths = expand_ref(
+        str(config_path),
+        config_data,
+        [str(workflow_step_path), str(workflow_step_path.parent / ".snappy_pipeline")],
+    )
+    return config, lookup_paths, config_paths
```
Nicolai-vKuegelgen, line 94: The `config` object returned as first element of the retruned tuple here is a string. However, most functions working with this expect it to be a (yaml representing) dictionary, likely objects are switched?

cubi_tk/snappy/common.py
```diff
@@ -37,6 +23,124 @@ class CouldNotFindBioMedSheet(Exception):
...
+
+    :param snappy_root_dir: Path to snappy root directory.
+    :type snappy_root_dir: str, pathlib.Path
+
+    :return: Returns loaded snappy configuration.
+    """
+    snappy_config = snappy_root_dir / ".snappy_pipeline" / "config.yaml"
+    with open(snappy_config, "r") as stream:
+        config = yaml.safe_load(stream)
+    return config
```
Nicolai-vKuegelgen, line 108: I can not find any references to this function?

cubi_tk/snappy/common.py
```diff
@@ -37,6 +23,124 @@ class CouldNotFindBioMedSheet(Exception):
...
+    def _catchall(self, *_, **__):
+        pass
+
+
+def get_available_snappy_workflows():
+    """Get all available snappy workflows in snappy_pipeline.
+
+    :return: List of ModuleInfo with individual workflow modules.
+    """
+    return list(pkgutil.iter_modules(snappy_workflows.__path__))
```
Nicolai-vKuegelgen, line 44: Snappy (in apps / snappy_snake) contains the `STEP_TO_MODULE` dict, that lists all fully implemented steps (and omits other workflow objects that might be under development or for internal use only), why not just import that?

cubi_tk/snappy/common.py
```diff
@@ -37,6 +23,124 @@ class CouldNotFindBioMedSheet(Exception):
...
+
+    config, lookup_paths, config_paths = load_workflow_step_configuration(workflow_step_path)
+    if config is None:
+        raise RuntimeError(f"Could not load workflow step confiuration for {workflow_step_path}")
+
+    step_name = config["pipeline_step"]["name"]
+
+    step_module_paths = get_available_snappy_workflow_paths()
+    module_path = step_module_paths[step_name]
+    module_config_path = module_path / "Snakefile"
```
Nicolai-vKuegelgen, line 128: Snappy already provides a function for this: snappy-pipeline / workflows / base . snakefile_path(step_name)

cubi_tk/snappy/common.py
```diff
@@ -37,6 +23,124 @@ class CouldNotFindBioMedSheet(Exception):
...
+    if obj_name is None:
+        raise RuntimeError(f"Could not find workflow object for {step_name}")
+
+    workflow_module = importlib.import_module(f".{step_name}", "snappy_pipeline.workflows")
+    workflow_class = getattr(workflow_module, obj_name)
+
+    workflow_object = workflow_class(
+        DummyWorkflow(), config, lookup_paths, config_paths, str(workflow_step_path)
+    )
+    dependencies = workflow_object.sub_workflows.keys()
```
Nicolai-vKuegelgen, line 140: Wouldn't it be better to use `workflow_object.previous_steps` here?

Also the 'workflow_object' should probably be name something like 'snappy_step_obejct' instead

cubi_tk/snappy/common.py
```diff
@@ -37,6 +23,124 @@ class CouldNotFindBioMedSheet(Exception):
...
+    step_module_paths = get_available_snappy_workflow_paths()
+    module_path = step_module_paths[step_name]
+    module_config_path = module_path / "Snakefile"
+    # get the name of the workflow step class name
+    obj_name = get_workflow_snakefile_object_name(module_config_path)
+    if obj_name is None:
+        raise RuntimeError(f"Could not find workflow object for {step_name}")
+
+    workflow_module = importlib.import_module(f".{step_name}", "snappy_pipeline.workflows")
+    workflow_class = getattr(workflow_module, obj_name)
```
Nicolai-vKuegelgen, line 135: I'm not really happy with having to parse the snappy source code to get the name of the workflow/Step-Manager object here.

From what I can see these objects are almost always the Camel-Case (+Workflow) of the step_name. Maybe we can rely on that? Neither is really a neat solution though.
Either way we should assure that the `name` attribute of the imported class matches the desired step_name.

cubi_tk/snappy/kickoff.py
```diff
@@ -19,39 +19,45 @@ def run(
...
-    # TODO: this assumes standard naming which is a limitation...
-    logger.info("Looking for pipeline directories (assuming standard naming)...")
+    logger.info("Looking for pipeline directories (needs to contain snappy config.yaml)...")
     logger.debug("Looking in %s", path)
-    step_set = {name for name in common.DEPENDENCIES if (path / name).exists()}
+    step_set = {}
+    folder_steps = common.get_snappy_step_directories(path)
+    for step_name, step_path in folder_steps.items():
+        dependencies = common.get_workflow_step_dependencies(step_path)
+        step_set[step_name] = dependencies
```
Nicolai-vKuegelgen, line 32: Why is there no check to ensure that directories with step names actually exist?
From what I can see this is also not checked by common.get_snappy_step_directories(path).

It seems that from here on `step_set`is used both for the steps that actually need to be run *and* for tracking the dependencies between the steps. I think those two things should be tracked separately, i.e. there should be a 'dependencies' dictionary.

cubi_tk/snappy/common.py
Nicolai-vKuegelgen, line 1: There number of functions added here is pretty high and almost all of them are used only to obtain the dependencies of the snappy steps. I think it might be worthwhile to collect all these functions into a class (or otherwise collect them, maybe some can be merged if they are only used in a single place?).

There also seem to be some functions that are not used at all?

Nicolai-vKuegelgen (changes requested):
Most important general points, rest is in details:
1) I don't think we should add the snappy code as a general requirement/dependency of cubi-tk, this can (or maybe should be) discussed though. I think checking for the possibility to import snappy_pipeline and then using that is reasonable though.
2) Several of the new functions defined in cubi-tk/snappy/common.py can probably be replaced by importing things already defined in snappy_pipeline. Doing this should also reduce the amount of new - and not test-covered - code. Also it might make sense to bundle all of the functionality to interface with the snappy code into a single class.

README.md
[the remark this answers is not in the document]
xiamaz: Given that `cubi-tk snappy` greatly depends on certain `snappy` internals (such as config file format and snappy pipeline steps). I believe cubi-tk should definitely depend on snappy.

The current solution still does not version snappy-pipeline, but does not break CI and users installing cubi-tk will automatically pull in snappy-pipeline as a dependency. I don't see this as more annoying than any other dependency.

Still I can see your point, given that `cubi-tk` can be used in a large number of scenarios. I would suggest adding snappy-pipeline as an optional dependency: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

This pattern is well known in many larger python projects and would allow features to be selectively included.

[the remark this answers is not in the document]
ericblanc20: I am with Nicolai here: I don't think `cubi-tk` should be depending on `snappy`. It is used by `sea-snap`, and possibly in the future by the single cell pipeline, for proteomics data, ...

I understand that in principle it is just one more dependency, but that dependency include `snakemake`, which has a ton of extra dependencies itself. I can see version conflicts becoming a problem, especially if `cubi-tk` is extended to other modalities.

xiamaz:
@Nicolai-vKuegelgen All comments have been adressed. Import is now optional and the logic for optional imports is also tested.

cubi_tk/snappy/snappy_workflows.py
```diff
@@ -0,0 +1,156 @@
...
+            raise RuntimeError(f"Could not find workflow object for {step_name}")
+
+        workflow_module = importlib.import_module(f".{step_name}", "snappy_pipeline.workflows")
+        workflow_class = getattr(workflow_module, obj_name)
+        assert workflow_class.name == step_name
+
+        workflow_object = workflow_class(
+            DummyWorkflow(), config, lookup_paths, config_paths, str(workflow_step_path)
+        )
+        dependencies = workflow_object.sub_workflows.keys()
```
Nicolai-vKuegelgen, line 155: I'm unsure if this is how snappy 'intends' to define dependencies, since there also should be `workflow_object.previous_steps`. Likely is either is fine though.

cubi_tk/snappy/kickoff.py
```diff
@@ -12,46 +12,63 @@
...
-    # TODO: this assumes standard naming which is a limitation...
-    logger.info("Looking for pipeline directories (assuming standard naming)...")
+    logger.info("Looking for pipeline directories (needs to contain snappy config.yaml)...")
     logger.debug("Looking in %s", path)
-    step_set = {name for name in common.DEPENDENCIES if (path / name).exists()}
+
+    manager = SnappyWorkflowManager.from_snappy()
+
+    if manager is None:
+        return None
```
Nicolai-vKuegelgen, line 35: I think it would also be okay to raise an error (with a nice message) here, since the `kickoff` function can rightfully depend on snappy installation. The warning given by the manager also conveys the necessray information though.

cubi_tk/snappy/kickoff.py
```diff
@@ -12,46 +12,63 @@
...
+
+    step_dependencies = {}
+    folder_steps = manager.get_snappy_step_directories(path)
+    for step_name, step_path in folder_steps.items():
+        dependencies = manager.get_workflow_step_dependencies(step_path)
+        assert all(dep in folder_steps for dep in dependencies)
```
Nicolai-vKuegelgen, line 41: It might better to raise a custom error here, since an AssertionError will not clearly communicate that there are missing dependecies in the snappy steps (could happen if i.e. the pipeline gets a major update and one returns to an old project).

tests/test_snappy_workflows.py
Nicolai-vKuegelgen, line 1: One could probably add further tests to check:
- if some steps (i.e. ngs_mapping) have no dependencies
- other steps have the correct dependencies (i.e. variant_calling should always depend on ngs_mapping)
- a dummy folder+config structure is properly parsed (this might need a lot / too much work though)

Nicolai-vKuegelgen (approved):
I think this looks good overall.

A few comments on things that I think could be further improved, but nothing that is absolutely needed.

PR #176, opened 2023-06-05
feat: allow for existing files when using irods download

### LINKED ISSUE #175: Allow pull-raw-data to work without force-overwriting data
Currently snappy pull-raw-data either requires force-overwriting of existing samples or requires samples to not exist. Without force-overwriting. Existing samples should be taken as-is and not redownloaded.

Issue an info if an existing file is used instead of a download.

xiamaz:
This fixes #175 by first comparing the filepath and only issuing the irods get if the file does not already exists.

# Limitations

This does not perform a md5 checksum

# Implementation details

Irods pythonclient directly issues a download from the get command if a localpath is included:   [ def get(self, path, local_path = None, num_threads = DEFAULT_NUMBER_OF_THREADS, **options):](https://github.com/irods/python-irodsclient/blob/bb03bb5bccf650838ae6468e6f90c60325a07997/irods/manager/data_object_manager.py#L111)

This unfortunately does not perform a md5 check against an existing file and instead only offers clobbering or fails with an exception.

## Possible alternative implementations

We can also handle this in the exception handler and trust the irods library to raise the correct expection. This would be more dependent on the specific irods library logic.

### DIFF
--- a/cubi_tk/snappy/pull_data_common.py
+++ b/cubi_tk/snappy/pull_data_common.py
@@ -1,6 +1,5 @@
-import os
-
 from datetime import datetime
+import os
 from pathlib import Path
 from types import SimpleNamespace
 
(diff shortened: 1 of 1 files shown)


xiamaz:
@holtgrewe Small convenience patch for cubi-tk file downloads. Should in most cases be expected instead of surprising behavior that downloaded files are not overwritten, if a `force_overwrite` flag exists in cubi-tk

sellth:
If you want to implement a md5 check, the get command can be used to access the hashsum stored in iRODS.

PR #180, opened 2023-06-08
fix: actually match files recursively

sellth:
Added missing parameter `recursive=True` as described here: https://docs.python.org/3/library/glob.html


Nicolai-vKuegelgen (approved):
LGTM

PR #181, opened 2023-06-08
feat: validate ISA-tabs after cookiecutter run

### LINKED ISSUE #178: Validate output of `cubi-tk isa-tpl`
As cookiecutter templates allow for invalid ISA-tab generation if user input does not follow spec (e. g. length mismatch in *Study Factors* and *Study Factor Types*) we should validate output before exiting. This mirrors what SODAR does when creating ISA-tabs using the templates.

sellth:
fixes #178 

Edit: Since switching to https://github.com/bihealth/cubi-isa-templates all templates are checked for validity on that side. I still think this feature can be useful in some cases.


cubi_tk/isa_tpl/__init__.py
```diff
@@ -255,4 +258,21 @@ def run(args, parser, subparser):  # pragma: nocover
     if not args.tpl:  # pragma: nocover
         return run_nocmd(args, parser, subparser)
     else:
-        return args.isa_tpl_cmd(args, parser, subparser)
+        args.isa_tpl_cmd(args, parser, subparser)
```
Nicolai-vKuegelgen, line 261: If running this would retrun an error (code) then it makes no sense to run the altamisa check, does it?

cubi_tk/isa_tpl/__init__.py
```diff
@@ -255,4 +258,21 @@ def run(args, parser, subparser):  # pragma: nocover
...
+            try:
+                with i.open() as i_file:
+                    args.input_investigation_file = i_file
+                    altamisa.apps.isatab_validate.run(args)
+            except (
+                altamisa.exceptions.ParseIsatabException,
+                altamisa.exceptions.CriticalIsaValidationWarning,
+            ):
+                shutil.rmtree(args.output_dir)
+                raise
```
Nicolai-vKuegelgen, line 278: Additionally this function now does not seem to have return value anymore? I'm not sure if that can lead to errors with the underlying code or if its 'only' a consistency issue, but either way this function should probably still have a return value (iirc cubi-tk simply retruns False/True or 0/1 as error codes).

Nicolai-vKuegelgen (changes requested):
generally looks good, but the current change removes retrun value from the changed function.

Issue #186, opened 2023-07-07
Sub command "snappy pull-sheets" broken with latest DNA sequencing ISA-tab template


holtgrewe:
**Describe the bug**
The latest DNA sequencing ISA-tab sample sheet does not use the non-standard column "Library Name" any more but rather uses the standard "Extract Name" now.

**To Reproduce**
1. Use latest DNA sequencing ISA-tab template, e.g., by using the wizard in SODAR
2. Try to pull the sheets with CUBI-tk
3. See "key not found" error

**Expected behavior**
We should interpret both the legacy and the new format.

**Additional context**
Here is the PR that has the change: bihealth/cubi-isa-templates#3

PR #187, opened 2023-07-07
fix: adjust to change in ISA-tab template (#186)

holtgrewe:
CC @ericblanc20 @Nicolai-vKuegelgen @xiamaz @mikkonie


xiamaz (approved):
LGTM

PR #184, opened 2023-06-28
feat: Fast irods querying

xiamaz:
Still work in progress but already reduces irods pull commands to <30s. Querying all files from a single sodar assay takes ~2.5s. The slow part is reading each `md5` checksum.

# Todo

- [x] clean up codebase
- [x] make filtering step cleaner
- [x] clean up checksum extraction

Since irods already offers native checksumming, its unclear whether the current md5 checksums are really necessary. This should be cleared up.

PR #188, opened 2023-07-11
fix: Fixes for changes introdueced when switching pull functions to query

Nicolai-vKuegelgen:
The previous changed (PR #184 ) had some overlooked bugs, this should fix them.


xiamaz (approved):
LGTM

PR #195, opened 2023-09-22
fix: Update ISA-tpl dependencies

sellth:
v0.1 of cubi-isa-templates contained a workaround for NCIT accession IDs which is no longer needed and in fact clashes with the fix deployed in the latest SODAR version.
 
Also, Cookiecutter fixed a display bug: https://github.com/cookiecutter/cookiecutter/issues/1939


ericblanc20 (approved):
Looks good to me...

PR #199, opened 2023-11-07
feat: Add generic SODAR ingest command

sellth:
re do of #168 

This adds a generic `sodar ingest` command which tries to mimic the base functionality of `irsync`, boosted by the power of using SODAR's API. Features include:

- no dependency on icommands; uses `iinit` session file if present, asks for password otherwise
- multiple source locations, single target collection
- recursive file matching
- creation of matched sub-folders in iRODS
- destination either as iRODS path or SODAR LZ UUID
- uses SODAR API for UUID destination & for querying LZ collections
- computes local checksums if missing
- triggers remote iRODS checksum computation
- scriptable; all user input can be skipped
- has a cool progress bar

- [x] rebase after PR #202 is merged.


cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,297 @@
...
+            self.lz_irods_path = self.args.destination  # pragma: no cover
+
+        # Build file list
+        source_paths = self.build_file_list()
+        if len(source_paths) == 0:
+            logger.info("Nothing to do. Quitting.")
+            sys.exit(0)
+
+        # Initiate iRODS session
+        irods_session = iRODSCommon().session
```
Nicolai-vKuegelgen, line 161: Will this be the same session as used for uploads? (since it's created by a different iRODScommon class instance)

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,297 @@
...
+        if not collections:
+            self.target_coll = self.lz_irods_path
+            logger.info("No subcollections found. Moving on.")
+        elif self.args.collection is None:
+            user_input = ""
+            input_valid = False
+            input_message = "####################\nPlease choose target collection:\n"
+            for index, item in enumerate(collections):
+                input_message += f"{index+1}) {item}\n"
+            input_message += "Select by number: "
```
Nicolai-vKuegelgen, line 187: This part sort of ignores the "--yes" flag (in that it still asks for user input even with --yes), maybe add a note to the cmd-line flag help strings about this?
Could also add a check_args function that gives error if --yes is set without a set collection

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,297 @@
...
+            self.target_coll = f"{self.lz_irods_path}/{collections[user_input - 1]}"
+
+        elif self.args.collection in collections:
+            self.target_coll = f"{self.lz_irods_path}/{self.args.collection}"
+        else:  # pragma: no cover
+            logger.error("Selected target collection does not exist in landing zone.")
+            sys.exit(1)
+
+        # Build transfer jobs and add missing md5 files
+        jobs = self.build_jobs(source_paths)
```
Nicolai-vKuegelgen, line 205: Same as above, this should use the same function names across cubi-tk

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,297 @@
...
+                self.lz_irods_path = lz_info.irods_path
+                logger.info(f"Target iRods path: {self.lz_irods_path}")
+            else:
+                logger.error("Target landing zone is not ACTIVE.")
+                sys.exit(1)
+        else:
+            self.lz_irods_path = self.args.destination  # pragma: no cover
+
+        # Build file list
+        source_paths = self.build_file_list()
```
Nicolai-vKuegelgen, line 155: I think it would be nice if all ingest/itransfer functions can use the same function namesin places like this.
However, this will likely need changes across cubi-tk and we need to agree on a standard

cubi_tk/sodar/ingest.py
```diff
@@ -0,0 +1,297 @@
...
+        for p in source_paths:
+            path_remote = f"{self.target_coll}/{str(p['ipath'])}"
+            md5_path = p["spath"].parent / (p["spath"].name + ".md5")
+
+            if md5_path.exists():
+                logger.info(f"Found md5 hash on disk for {p['spath']}")
+            else:
+                md5sum = compute_md5_checksum(p["spath"])
+                with md5_path.open("w", encoding="utf-8") as f:
+                    f.write(f"{md5sum}  {p['spath'].name}")
```
Nicolai-vKuegelgen, line 276: md5 computation could be made controllable by user (with default = on)

cubi_tk/irods_common.py
```diff
@@ -0,0 +1,250 @@
...
+                        if sync and session.data_objects.exists(job.path_remote):
+                            t.update(job.bytes)
+                            continue
+                        session.data_objects.put(job.path_local, job.path_remote)
+                        t.update(job.bytes)
+                except Exception as e:  # pragma: no cover
+                    logger.error(f"Problem during transfer of {job.path_local}")
+                    logger.error(self.get_irods_error(e))
+                    sys.exit(1)
+            t.clear()
```
Nicolai-vKuegelgen, line 202: The put function could maybe always trigger a self.chksum() after (successful) completion?
(At least as an option)

Nicolai-vKuegelgen (commented):
Looks very good overall!

Most comments are only minor questions / suggestions.

If you try to imitate irsync functionality why not name they whole thing `cubi-tk sodar irsync` ?
Ah, and: it doesn't fully mimic irsync does it? Because one can not give a full irods path, instead only LZ path & collection sepeartely

sellth:
Thanks for the review. Hopefully this will look a lot less messy after we merge #199.
 
> If you try to imitate irsync functionality why not name they whole thing `cubi-tk sodar irsync` ?

irsync implies that it goes both ways, this is only an uploader. It also implements a lot of SODAR-specific quality of life features such as LZ query and automatic .md5 calculation/handling.

> Ah, and: it doesn't fully mimic irsync does it? Because one can not give a full irods path, instead only LZ path & collection sepeartely

You actually can give it a full iRODS path. It will skip the subcollection selection of there are no subcollections found.

Will look at the rest of the comments now.

cubi_tk/irods_common.py
```diff
@@ -0,0 +1,250 @@
...
+                file_log.set_description_str(
+                    f"File [{n + 1}/{len(self.__jobs)}]: {Path(job.path_local).name}"
+                )
+                try:
+                    with self.session as session:
+                        if recursive:
+                            self._create_collections(job)
+                        if sync and session.data_objects.exists(job.path_remote):
+                            t.update(job.bytes)
+                            continue
```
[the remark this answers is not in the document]
sellth: Technically this is part of #202. Comparing checksums, while obviously more precise, has some caveats:
- Some files might not have server-side checksums yet.
- Remote checksum computation takes time which might not be what the user wants.
- It breaks with the established SODAR workflow of Upload > Checksum > Validate

That's why I opted for the simple "does file exist" check.

cubi_tk/sodar/ingest.py
sellth: It's not the same session. Uploads and checksumming are internally handled by the iRODSTransfer class. This session is used to query LZ subcollections.

cubi_tk/irods_common.py
sellth: Technically part of #202. I would leave this as is in the iRODSTransfer class. Ingest functions can automate the checksumming process or put it behind a -K flag. I would argue it's not really critical any more since SODAR calculates server-side checksums on LZ validation.

cubi_tk/irods_common.py
```diff
@@ -0,0 +1,250 @@
...
+
+        if isinstance(token, list) and token:
+            irods_auth_path.write_text(encode(token[0]))
+            irods_auth_path.chmod(0o600)
+        else:
+            logger.warning("No token found to be saved.")
+
+    @property
+    def session(self):
+        return self._init_irods()
```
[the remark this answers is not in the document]
sellth: Two things to consider:
- One session object for everything could time out for very long transfers.
- The drawback is a slight time cost, but considering the size of our usual files, that's negligible.

cubi_tk/sodar/ingest.py
sellth: Why would the user want to turn it off?

cubi_tk/sodar/ingest.py
sellth: This is the same name as ingest_fastq uses.

cubi_tk/sodar/ingest.py
sellth: Maybe, even though not much is gained from this if you ask me. I would however discuss this at a later time.

cubi_tk/sodar/ingest.py
sellth: help text added

cubi_tk/sodar/ingest.py
Nicolai-vKuegelgen: One might expect to have all md5 files already and not want to upload 'incomplete' file sets

cubi_tk/irods_common.py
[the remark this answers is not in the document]
Nicolai-vKuegelgen: mmh yeah, this is a bit tricky I guess.
Lets leave it as is and try to remember to look back if we ever have runtime issues (maybe leave a comment in the code?)

cubi_tk/irods_common.py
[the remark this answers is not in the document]
Nicolai-vKuegelgen: I thought one would do the checksum check after the file exists check (i.e. an "am I changing the remote existing file" check). But if the remote file doesn't have checksums yet then this is indeed a bit tricky.
Also doesn't really need to be solved right now

cubi_tk/sodar/ingest.py
Nicolai-vKuegelgen: Yeah this needs to be discussed overall (and probably also needs to be done on all functions at the same time)

cubi_tk/sodar/ingest.py
sellth: Still not getting it. What kind of incomplete file sets?

Nicolai-vKuegelgen (approved):
LGTM

PR #208, opened 2024-01-04
refactor: Use iRODSCommon as parent class

sellth:
Utilise common class introduced in #202 for `irods check` command.


Nicolai-vKuegelgen (commented):
Looks good to me so far.

Do you want to merge this as is, or refactor all the other functions that could/should this this in one go?

PR #217, opened 2024-01-24
feat: Switching `cubi-tk sodar ingest-fastq` from icommands to irods_common

Nicolai-vKuegelgen:
Note: in the future the `SnappyItransferCommandBase` class should be changed to directly use the iRODSCommon functions for uplods, however that will need more work, so I"m only doing the necessary changes here for now.


sellth (approved):
Looks good to me; one small comment.

sellth (approved):
Looks good to me; one small comment.

PR #224, opened 2024-02-08
feat: allow varfish case resubmission

Nicolai-vKuegelgen:
Previous varfish-cli versions always allowed case resubmission, the new 0.6.2 version does not do this by default, so we need the extra option in cubi-tk.

(I'm not sure why git thinks these other commits are part of this, they are already merged into main as of #222 )

### DIFF
--- a/cubi_tk/snappy/varfish_upload.py
+++ b/cubi_tk/snappy/varfish_upload.py
@@ -4,11 +4,11 @@ import argparse
 import glob
 import os
 import pathlib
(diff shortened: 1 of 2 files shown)


sellth:
I can answer the git question right away, because these commits were squashed into cb10dbbcedd9e9d287428f14873420c5b6d4898c. This PR will also be squashed into a new commit, so we're good.

sellth (approved):
lgtm

PR #227, opened 2024-03-11
feat: irods download refactoring and new generic sodar downloader with preset for dragen data (#226 )

### LINKED ISSUE #201: Refactor commonly use functionality, esp. from `cubi-tk snappy`
Several cubi-tk functions that are used/usable by all modules are currently part of the `cubi-tk snappy` subcollection, but would be better placed in general or common files. Specifically these are: 
- `itransfer_common.py` which contains general functions for data upload to irods (but inherits some functionality from snappy specific sample sheet prasing)
- `pull_data_common.py` which contains general functions for data download and can probably be moved without many changes, except that it inherits from `IrodsCheckCommand` from the irods module, which should probably also be refactored
- `retrieve_irods_collection.py` which contains general functions for collecting information about files in irods and can probably be moved without many changes
 
All functionality that is commonly used by cubi-tk for interfacing with irods &/ Sodar should be refactored into files in the common space (i.e. irods_common / ... ) 

**Describe alternatives you've considered**
NA / status quo

**Additional context**
NA

### LINKED ISSUE #226: New sodar downloader for Dragen data
The Dragen/Sodar/.../Varfush workflows need specific sets of files from Sodar (including different file types). It would be good to have a specific downloader for this.

A function that can download a set files based on specific (remote) file patterns.
Ideally, one that comes with the option/technology to easily built other bulk downloaders.

**Describe alternatives you've considered**
N/A

**Additional context**
A new downloader class / function should directly use the new iRODSCommon classes (#208 )

Nicolai-vKuegelgen:
Fixes #226 
Replaces #208 

Partially fixes #201

### DIFF
--- a/cubi_tk/irods_common.py
+++ b/cubi_tk/irods_common.py
@@ -267,6 +267,7 @@ class iRODSTransfer(iRODSCommon):
                     )
                     continue
                 try:
+                    Path(job.path_local).parent.mkdir(parents=True, exist_ok=True)
                     with self.session as session:
                         session.data_objects.get(job.path_remote, job.path_local, **kw_options)
                     t.update(job.bytes)
--- a/cubi_tk/irods_common.py
+++ b/cubi_tk/irods_common.py
@@ -295,7 +295,7 @@ class iRODSRetrieveCollection(iRODSCommon):
         :param irods_env_path: Path to irods_environment.json
         :type irods_env_path: pathlib.Path, optional
         """
-        super.__init__(ask, irods_env_path)
+        super().__init__(ask, irods_env_path)
         self.hash_scheme = hash_scheme
 
     def retrieve_irods_data_objects(self, irods_path: str) -> Dict[str, List[iRODSDataObject]]:
@@ -308,15 +308,15 @@ class iRODSRetrieveCollection(iRODSCommon):
         """
 
         # Connect to iRODS
-        with self.session as irods_session:
+        with self.session as session:
             try:
-                root_coll = irods_session.collections.get(irods_path)
+                root_coll = session.collections.get(irods_path)
 
                 # Get files and run checks
                 logger.info("Querying for data objects")
 
                 if root_coll is not None:
-                    irods_data_objs = self.get_data_objs(root_coll)
+                    irods_data_objs = self._irods_query(session, root_coll)
                     irods_obj_dict = self.parse_irods_collection(irods_data_objs)
                     return irods_obj_dict
 
@@ -326,24 +326,27 @@ class iRODSRetrieveCollection(iRODSCommon):
 
         return {}
 
-    def get_data_objs(
-        self, root_coll: iRODSCollection
+    def _irods_query(
+        self,
+        session: iRODSSession,
+        root_coll: iRODSCollection,
     ) -> Dict[str, Union[Dict[str, iRODSDataObject], List[iRODSDataObject]]]:
         """Get data objects recursively under the given iRODS path."""
-        data_objs = dict(files=[], checksums={})
-        ignore_schemes = [k.lower() for k in HASH_SCHEMES if k != self.hash_scheme.upper()]
-        irods_sess = root_coll.manager.sess
 
-        query = irods_sess.query(DataObjectModel, CollectionModel).filter(
+        ignore_schemes = [k.lower() for k in HASH_SCHEMES if k != self.hash_scheme.upper()]
+
+        query = session.query(DataObjectModel, CollectionModel).filter(
             Like(CollectionModel.name, f"{root_coll.path}%")
         )
 
+        data_objs = dict(files=[], checksums={})
         for res in query:
-            # If the 'res' dict is not split into Colllection&Object the resulting iRODSDataObject is not fully functional, likely because a name/path/... attribute is overwritten somewhere
+            # If the 'res' dict is not split into Colllection&Object the resulting iRODSDataObject is not fully functional,
+            # likely because a name/path/... attribute is overwritten somewhere
             coll_res = {k: v for k, v in res.items() if k.icat_id >= 500}
             obj_res = {k: v for k, v in res.items() if k.icat_id < 500}
             coll = iRODSCollection(root_coll.manager, coll_res)
-            obj = iRODSDataObject(irods_sess.data_objects, parent=coll, results=[obj_res])
+            obj = iRODSDataObject(session.data_objects, parent=coll, results=[obj_res])
 
             if obj.path.endswith("." + self.hash_scheme.lower()):
                 data_objs["checksums"][obj.path] = obj
--- a/cubi_tk/irods_common.py
+++ b/cubi_tk/irods_common.py
@@ -261,7 +261,7 @@ class iRODSTransfer(iRODSCommon):
                 file_log.set_description_str(
                     f"File [{n + 1}/{len(self.__jobs)}]: {Path(job.path_local).name}"
                 )
-                if os.path.exists(job.path_local) and not force_overwrite:
+                if os.path.exists(job.path_local) and not force_overwrite:  # pragma: no cover
                     logger.info(
                         f"{Path(job.path_local).name} already exists. Skipping, use force_overwrite to re-download."
                     )
@@ -320,7 +320,7 @@ class iRODSRetrieveCollection(iRODSCommon):
                     irods_obj_dict = self.parse_irods_collection(irods_data_objs)
                     return irods_obj_dict
 
-            except Exception as e:
+            except Exception as e:  # pragma: no cover
                 logger.error("Failed to retrieve iRODS path: %s", self.get_irods_error(e))
                 raise
 
--- a/cubi_tk/irods_common.py
+++ b/cubi_tk/irods_common.py
@@ -1,15 +1,23 @@
+from collections import defaultdict
 import getpass
 import os.path
 from pathlib import Path
-from typing import Iterable
+import re
+from typing import Dict, Iterable, List, Union
 
 import attrs
+from irods.collection import iRODSCollection
+from irods.column import Like
+from irods.data_object import iRODSDataObject
 from irods.exception import (
     CAT_INVALID_AUTHENTICATION,
     CAT_INVALID_USER,
     CAT_PASSWORD_EXPIRED,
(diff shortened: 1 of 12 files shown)


cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,367 @@
...
+
+    command_name = "pull-data-collection"
+
+    presets = {
+        "dragen": [
+            "**/*_FAM_dragen.fam.hard-filtered.vcf.gz"
+            "**/*_FAM_dragen.fam.hard-filtered.vcf.gz.tbi",
+            "**/*.qc-coverage*.csv",
+            "**/*.ped",
+            "**/*.mapping_metrics.csv",
```
xiamaz, line 31: aren't we missing the sv files?

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,367 @@
...
+        iRODSTransfer(transfer_jobs).get(self.args.overwrite)
+
+        logger.info("All done. Have a nice day!")
+        return 0
+
+    @staticmethod
+    def parse_sample_tsv(tsv_path, sample_col=1, n_header_cols=1, skip_comments=True) -> List[str]:
+        samples = []
+        with open(tsv_path, "r") as f:
+            for _ in range(n_header_cols):
```
xiamaz, line 251: I would prefer using csv library or any library that properly supports the format we are actually working with

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,367 @@
...
+        for filename, irodsobjs in remote_files_dict.items():
+            for irodsobj in irodsobjs:
+                # Path needs to be stripped down to collections (=remove assay part & upwards)
+                path = PurePosixPath(irodsobj.path).relative_to(PurePosixPath(common_assay_path))
+
+                collection = path.parts[0]
+
+                # Check if collection (=1st element of striped path) matches any of the samples
+                if samples and not substring_match:
+                    sample_match = any([s.lower() == collection.lower() for s in samples])
```
xiamaz, line 310: why are matches case insensitive? is irods actually case insensitive?

xiamaz (changes requested):
Some comments from my side

cubi_tk/sodar/pull_data_collection.py
Nicolai-vKuegelgen: correct, adding them

cubi_tk/sodar/pull_data_collection.py
Nicolai-vKuegelgen: mmh, it seems the csv library doesn't support skipping commented lines (which we should have), so I don't think it's really woth using that.
We could go for pandas (which is already a cubi-tk dependency anyway), I'll look into that

cubi_tk/sodar/pull_data_collection.py
Nicolai-vKuegelgen: mmh, I thought I'd taken that from somewhere else in cubi-tk but that doesn't seem to be the case. We can go case-sensitive I guess

cubi_tk/sodar/pull_data_collection.py
Nicolai-vKuegelgen: Switched to pandas for tsv parsing

xiamaz (approved):
lgtm

cubi_tk/irods_common.py
```diff
@@ -242,13 +261,115 @@ def get(self):
...
+        query = session.query(DataObjectModel, CollectionModel).filter(
+            Like(CollectionModel.name, f"{root_coll.path}%")
+        )
+
+        data_objs = dict(files=[], checksums={})
+        for res in query:
+            # If the 'res' dict is not split into Colllection&Object the resulting iRODSDataObject is not fully functional,
+            # likely because a name/path/... attribute is overwritten somewhere
+            coll_res = {k: v for k, v in res.items() if k.icat_id >= 500}
+            obj_res = {k: v for k, v in res.items() if k.icat_id < 500}
```
tedil, line 348: Perhaps store the `500` in some sensibly named constant, otherwise seems a bit arbitrary (at least to me, as someone who is not familiar with irods).

cubi_tk/irods_common.py
```diff
@@ -242,13 +261,115 @@ def get(self):
...
+            coll = iRODSCollection(root_coll.manager, coll_res)
+            obj = iRODSDataObject(session.data_objects, parent=coll, results=[obj_res])
+
+            if obj.path.endswith("." + self.hash_scheme.lower()):
+                data_objs["checksums"][obj.path] = obj
+            elif obj.path.split(".")[-1] not in ignore_schemes:
```
tedil, line 354: There's a `path.splitext`, but that includes the leading `.`, so perhaps not worth the hassle to change that here + changing `ignore_schemes`.

cubi_tk/irods_common.py
```diff
@@ -242,13 +261,115 @@ def get(self):
...
+                data_objs["files"].append(obj)
+
+        return data_objs
+
+    @staticmethod
+    def parse_irods_collection(irods_data_objs) -> Dict[str, List[iRODSDataObject]]:
```
tedil, line 360: I think the min requirement is python 3.10, so using the types directly should be fine:
```suggestion
    def parse_irods_collection(irods_data_objs) -> dict[str, list[iRODSDataObject]]:
```

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,364 @@
...
+        :param irods_env_path: Path to irods_environment.json
+        :type irods_env_path: pathlib.Path, optional
+        """
+        super().__init__()
+        #: Command line arguments.
+        self.args = args
+
+    @classmethod
+    def setup_argparse(cls, parser: argparse.ArgumentParser) -> None:
+        """Setup arguments for ``check-remote`` command."""
```
tedil, line 58:
```suggestion
        """Setup arguments for ``pull-data-collection`` command."""
```

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,364 @@
...
+
+        parser.add_argument(
+            "project_uuid",
+            help="SODAR project UUID",
+        )
+
+    @classmethod
+    def run(
+        cls, args, _parser: argparse.ArgumentParser, _subparser: argparse.ArgumentParser
+    ) -> typing.Optional[int]:
```
tedil, line 160: In python 3.10+, `type | None` is preferred over (and equivalent to) `Optional[type]` -- as far as I know. But that's cosmetic anyways.
```suggestion
    ) -> int | None:
```

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,364 @@
...
+
+        return res
+
+    def execute(self) -> typing.Optional[int]:
+        """Execute the transfer."""
+        res = self.check_args(self.args)
+        if res:  # pragma: nocover
+            return res
+
+        logger.info("Starting cubi-tk snappy pull-processed-data")
```
tedil, line 199: Is that the correct command name here?

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,364 @@
...
+
+        logger.info("All done. Have a nice day!")
+        return 0
+
+    @staticmethod
+    def parse_sample_tsv(tsv_path, sample_col=1, skip_rows=0, skip_comments=True) -> List[str]:
+        extra_args = {"comment": "#"} if skip_comments else {}
+        df = pd.read_csv(tsv_path, sep="\t", skiprows=skip_rows, **extra_args)
+        try:
+            samples = list(df.iloc[:, sample_col - 1])
```
tedil, line 257: Would it make sense to use a set instead?

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,364 @@
...
+        for filename, irodsobjs in remote_files_dict.items():
+            for irodsobj in irodsobjs:
+                # Path needs to be stripped down to collections (=remove assay part & upwards)
+                path = PurePosixPath(irodsobj.path).relative_to(PurePosixPath(common_assay_path))
+
+                collection = path.parts[0]
+
+                # Check if collection (=1st element of striped path) matches any of the samples
+                if samples and not substring_match:
+                    sample_match = any([s == collection for s in samples])
```
tedil, line 307:
```suggestion
                    sample_match = any(s == collection for s in samples)
```

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,364 @@
...
+                elif samples:
+                    sample_match = any([s in collection for s in samples])
```
tedil, line 309:
```suggestion
                    sample_match = any(s in collection for s in samples)
```

cubi_tk/sodar/pull_data_collection.py
```diff
@@ -0,0 +1,364 @@
...
+                else:
+                    sample_match = True
+
+                if not sample_match:
+                    continue
+
+                if file_patterns:
+                    file_pattern_match = any([p for p in file_patterns if path.match(p)])
```
tedil, line 317:
```suggestion
                    file_pattern_match = any(p for p in file_patterns if path.match(p))
```

tedil (commented):
I can't really comment on the logic here, as I am not particularly familiar with irods, so most of my comments are small-ish and optional suggestions (regarding type hints and generators instead of list comprehensions).
I didn't mark all type hints, but with python 3.10+ (which I think is the minimum requirement for cubi-tk), you can simply use the builtin directly, no need to import from typing for most of them.

cubi_tk/irods_common.py
Nicolai-vKuegelgen: 500 seems to be the magic cutoff betwenn object & collection ids in irods. I could not really find documentation for that, but giving it a name sounds good.

cubi_tk/irods_common.py
Nicolai-vKuegelgen: I think the "obj.path" is a string anyway, so I I'd rather not change too much here

PR #231, opened 2024-04-30
fix: parsing snappy config for  with WGS projects

Nicolai-vKuegelgen:


tedil (approved):
The comments still have the hyphen instead of the underscore, but other than that: LGTM

Issue #234, opened 2024-08-15
Not all landing zones listed with `cubi-tk sodar list-landingzones`


your-highness:
**Describe the bug**
When using `cubi-tk sodar landing-zone-list` only a subset of all landing zones is listed, _i.e._ only ones marked as "ACTIVE". Version 0.4.0

**To Reproduce**
1. Run `cubi-tk sodar landing-zone-list`
```bash
>cubi-tk sodar landing-zone-list f2acceb7-067d-41a4-8e39-236c022678f1
[I 240815 15:19:04] Starting cubi-tk sodar landing-zone-list
[I 240815 15:19:04]   args: Namespace(verbose=False, config=None, sodar_server_url=None, sodar_api_token='87a94aa011bf1d6a524f53f17e4d06f609e723e39a0e6e96d4d7181b329325cc', cmd='sodar', sodar_cmd=<bound method ListLandingZoneCommand.run of <class 'cubi_tk.sodar.lz_list.ListLandingZoneCommand'>>, sodar_url='https://sodar.bihealth.org/', unless_exists=False, dry_run=False, format_string=None, project_uuid='f2acceb7-067d-41a4-8e39-236c022678f1')
{"sodar_uuid": "509fb34a-3803-4015-a532-9105a40413b3", "date_modified": "2024-08-14T16:02:52.299700+02:00", "status": "ACTIVE", "project": "f2acceb7-067d-41a4-8e39-236c022678f1", "title": "20240813_182919", "description": "", "user": {"sodar_uuid": "e1441046-2727-41cb-a620-6453201f811e", "username": "svc-bih-lb-upload@CHARITE", "name": "svc-bih-lb-upload", "email": ""}, "assay": "928c4658-fa58-4486-b6ec-862911c160ab", "status_info": "Successfully validated 89 files", "configuration": null, "config_data": {}, "irods_path": "/sodarZone/projects/f2/f2acceb7-067d-41a4-8e39-236c022678f1/landing_zones/svc-bih-lb-upload@CHARITE/cads_exomes_diagnostics/cads_exomes_diagnostics_genome_sequencing_nucleotide_sequencing/20240813_182919"}
[...]

PR #239, opened 2024-09-06
feat: add `--filter-status` option for controlling which landing zones are listed

tedil:
TODO:

- [ ] find out what the status enum's variants are
- [ ] add test

### DIFF
--- a/src/cubi_tk/sodar/lz_list.py
+++ b/src/cubi_tk/sodar/lz_list.py
@@ -68,7 +68,7 @@ class ListLandingZoneCommand:
             default=set(),
             action="append",
             choices=["ACTIVE", "FAILED", "VALIDATING"],
-            help="Filter landing zones by status",
+            help="Filter landing zone by status",
         )
 
         parser.add_argument("project_uuid", help="UUID of project to create the landing zone in.")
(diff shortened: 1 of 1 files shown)


tedil:
This is one way to address #234, comments welcome

cubi_tk/sodar/lz_list.py
```diff
@@ -62,6 +62,16 @@ def setup_argparse(cls, parser: argparse.ArgumentParser) -> None:
             help="Format string for printing, e.g. %%(uuid)s",
         )
 
+        status = ["ACTIVE", "FAILED", "VALIDATING"]
```
Nicolai-vKuegelgen, line 65: These values should probably be defined elsewhere?
I'd say as class attributes, either of this class or in a class that directly interacts with the Sodar landingzone API. Ideally the possible values would even be derived from Sodar or its API somehow, but that might be tricky for use in argparse setup

tedil: Hm, they're at least not defined in the models in sodar-cli https://github.com/bihealth/sodar-cli/blob/a62505ff9b1365f150bce54c9b2b5e638f245f86/src/sodar_cli/api/models.py#L117

But I will at least move them somewhere else.

Nicolai-vKuegelgen (approved):
LGTM

PR #240, opened 2025-01-09
feat: add `cubi-tk sodar update-samplesheet`

Nicolai-vKuegelgen:
New function to allow flexible updates of ISA samplesheets. Also introduce a base class to implement SodarAPI features, as the currently used sodar_cli package is heavily outdated.
Features:
- can read ped files and/or command line defined sample specific (meta)data
- mapping of all ped/sampledata fields to ISA columns can be speciefied. Dynamic/derived columns are also possible 
- default settings for the ISA germline-DNA-sequencing sheet or the new Modellvorhaben project
- global metadata can be defined for all samples at once
- both adding new entries and updating entries is possible (--overwrite required for updating non-empty values)
- Values are mapped to the ISA table based on column name matching, parsing/validation of the ISA graph structure is NOT done by this function (but errors will be returned by the Sodar API anyway)

TBD:
- tsv or similar input for collected sampledata in another file format than ped

Note: this is a potential replacement/side-grade of `cubi-tk sodar add-ped`


cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+            )
+
+        # Base field name mapping (also used by ped)
+        sample_field_mapping = sheet_default_config[self.args.defaults]["field_column_mapping"]
+
+        # Sample data, if given, need to matche the defined fields (from default or sample_fields)
+        if self.args.sample_data:
+            if self.args.sample_fields:
+                sample_field_mapping.update(
+                    OrderedDict(
```
tedil, line 325: Which minium python version does cubi-tk require? Since python 3.7, dicts are guaranteed to be ordered by insertion order.
In that case, can probably be simplified to something along the lines of
`{k: (v or k) for (k, _, v) in map(lambda s: s.partition("="), self.args.sample_fields)}` (but not sure if that is more readable)

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+                    + ", ".join([f"{col} (from {field})" for field, col in unknown_fields])
+                )
+                raise NameError(msg)
+
+        return sample_field_mapping
+
+    def gather_ISA_column_names(
+        self, study: pd.DataFrame, assay: pd.DataFrame
+    ) -> dict[str, list[tuple[str, str]]]:
+        isa_regex = re.compile(r"(Characteristics|Parameter Value|Comment)\[(.*?)]")
```
tedil, line 364: Is the `[` escaped or do we match a literal `\` here?

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+            map(lambda x: orig_col_name(isa_regex.sub(r"\2", x)), study_cols + assay_cols)
+        )
+        isa_long_names = list(map(orig_col_name, study_cols + assay_cols))
+        isa_names_unique = study_cols + assay_cols
+        isa_table = ["study"] * len(study_cols) + ["assay"] * len(assay_cols)
+
+        out = defaultdict(list)
+        for short, long, uniq, table in zip(
+            isa_short_names, isa_long_names, isa_names_unique, isa_table
+        ):
```
tedil, line 378: Add `strict=True` to the arguments of `zip` to ensure they're all of the same length

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+
+    def collect_sample_data(
+        self,
+        isa_names: dict[str, list[tuple[str, str]]],
+        sample_field_mapping: dict[str, str],
+    ) -> pd.DataFrame:
+        ped_mapping = sheet_default_config[self.args.defaults]["ped_to_sampledata"]
+        if self.args.ped_field_mapping:
+            allowed_sample_col_values = [*sample_field_mapping] + [*isa_names]
+            for ped_col, sample_col in self.args.ped_field_mapping:
```
tedil, line 432: Not sure about the types here, the value is a str in case of sample_field_mapping but a list[...] in case of isa_names

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+
+                isa_table.loc[isa_col.index[empty_values], col] = (
+                    updates[col].loc[empty_values].values
+                )
+
+        # Check which cols should be autofilled ("Protocol REF" needs to be, for ISA to work)
+        autofill_cols = {
+            col: isa_table[col].unique()[0]
+            for col in isa_table.columns
+            if orig_col_name(col) == "Protocol REF"
```
tedil, line 605: So _only_ columns named "Protocol REF" will be autofilled?

cubi_tk/sodar_api.py
```diff
@@ -0,0 +1,151 @@
...
+                raise ParameterException(
+                    "SODAR API token not given on command line and not found in toml config files."
+                )
+            self.sodar_api_token = toml_config.get("global", {}).get("sodar_api_token")
+            if not self.sodar_api_token:
+                raise ParameterException(
+                    "SODAR API token not found in config files. Please specify on command line."
+                )
+        if not is_uuid(self.project_uuid):
+            logger.warning("Project UUID does not look like valid UUID.")
```
tedil, line 46: Would it make sense to error here instead of warn? If the project uuid is not a uuid, it will probably error downstream anyway, right?

cubi_tk/sodar_api.py
```diff
@@ -0,0 +1,151 @@
...
+        api: Literal["samplesheets", "landingzones"],
+        action: str,
+        method: Literal["get", "post"] = "get",
+        params: dict = None,
+        data: dict = None,
+        files: dict = None,
+    ) -> dict:
+        url = self.sodar_url + api + "/api/" + action + "/" + self.project_uuid
+        if params:
+            url += "?" + "&".join([f"{k}={v}" for k, v in params.items()])
```
tedil, line 90: Consider making use of urllib, this will also encode the url properly in case of weird characters etc.

tedil (changes requested):
I'm not too familiar with ISA, so most of my comments are more from a coding perspective and some general remarks.
I really like all the sanity checks and validations that you do here.

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+        Family-ID: Family
+        Analysis-ID: Source Name
+        Paternal-ID: Father
+        Maternal-ID: Mother
+        Sex: Sex
+        Phenotype: "Disease status"
+        Individual-ID: "Individual-ID"
+        Probe-ID: Probe-ID
+        Barcode: "Barcode sequence"
+        Barcode-Name: "Barcode name"
```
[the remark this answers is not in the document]
Nicolai-vKuegelgen: I inititally quote everything that had a space and the changed some things ...
I guess even with spaces the quotes aren't really necessary?

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+"""
+
+# ruamel round-trip loader uses ordered dicts
+yaml = YAML(typ="rt")
+sheet_default_config = yaml.load(sheet_default_config_yaml)
+
+
+def orig_col_name(col_name: str) -> str:
+    """Return the original column name without any suffixes."""
+    return re.sub(r"\.[0-9]+$", "", col_name)
```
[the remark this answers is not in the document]
Nicolai-vKuegelgen: ISAtab files tend to have some column names duplicated, the pandas parser adds suffixes to those.

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+            help="Do not automatically fill values for non-specified metadata columns that have a single unique value "
+            "in the existing samplesheet. Note: ontology terms & references will never be autofilled.",
+        )
+
+        parser.add_argument(
+            "--snappy-compatible",
+            action="store_true",
+            help="Transform IDs so they are compatible with snappy processing "
+            "(replaces '-' with '_' in required ISA fields).",
+        )
```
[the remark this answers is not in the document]
Nicolai-vKuegelgen: `snappy_pipeline` uses biomedsheets, which does indeed NOT like hyphens, the ISA files are translated into biomedsheets by other cubi-tk commands.
Changing biomedsheets is not so trivial iirc, another option would be to completely replace it and make snappy read sample info from some other (less restrictive) table format. Maybe using pep schemes?
Either way not a small change

cubi_tk/sodar/update_samplehseet.py
```diff
@@ -0,0 +1,628 @@
...
+        return ret
+
+    def parse_sampledata_args(self, isa_names) -> OrderedDict[str, str]:
+        """Build a dict of Sample-data filed name mappings and check consistency of provided data."""
+
+        # Some samples are defined, either through ped or sample data
+        if not self.args.ped and not self.args.sample_data:
+            raise ValueError(
+                "No sample data provided. Please provide either a ped file (`--ped`) or sample data (`-s`)."
+            )
```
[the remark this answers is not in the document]
Nicolai-vKuegelgen: yes using both is valid (and probably needs to be: ped files only contain certain datafields, but with the MV we need the ability to upload metadata for others as well. In order to still support ped files the remaining data needs to given through another option)

cubi_tk/sodar/update_samplehseet.py
Nicolai-vKuegelgen: ah I didn't know that about dictionaries. I think cubi-tk should have at least 3.7, so a few things here can probably be simplified

cubi_tk/sodar/update_samplehseet.py
Nicolai-vKuegelgen: It's escaped (pycharm complained if I also espaced the correspondign closing bracket)

cubi_tk/sodar/update_samplehseet.py
Nicolai-vKuegelgen: This *should* generate (addable) lists of the dict keys, which are both str. Is there a better way to achieve this?

cubi_tk/sodar/update_samplehseet.py
Nicolai-vKuegelgen: Yes, I initially wanted to (always) autofill some other columns as well, but that was able to break ISA compatibility.

cubi_tk/sodar_api.py
Nicolai-vKuegelgen: We can do that yes. testing will have to use an dummy UUID, but that's not a bad thing, I guess

cubi_tk/sodar/update_samplehseet.py
tedil: To get this straight, what is it supposed to match?

Nicolai-vKuegelgen: It (is supposed to) match the column name that is shown in Sodar, but in ISA is encoded as a specific column type.
So it looks something like this in the ISAtab file:
`Characteristics[Column name]`
Having users always give that full name would be a bit much, I think

cubi_tk/sodar/update_samplehseet.py
tedil: For me, I think doing `allowed_sample_col_values = sample_field_mapping.keys() | isa_names.keys()` would suffice to be a little bit more explicit/descriptive.
(also just noticed that this file is called `update_samplehseet`, we should fix that)

cubi_tk/sodar/update_samplehseet.py
tedil: Ah okay. Then this is a bit convoluted for the singular column that is auto filled, but might become useful in the future?

tests/test_sodar_update_samplesheet.py
```diff
@@ -0,0 +1,619 @@
...
+    USC = helper_update_UCS(arg_list, UCS_class_object)
+    assert USC.parse_sampledata_args(isa_names) == expected
+
+    # manually defined mapping
+    arg_list = [
+        "--sample-fields",
+        "Dummy-ID=Source Name",
+        "Sample Name",
+        "Extract Name",
+        "barcode=Barcode sequence",
```
tedil, line 201: I suspect this might behave unexpectedly depending on shell/quoting?

tedil: Ah, nvm, `arg_list` is forwarded as is to `argparse`.

tedil (approved):
LGTM, the only thing I'd like to see as a little bonus is an example in the documentation/manual or so.

tests/test_sodar_update_samplesheet.py
Nicolai-vKuegelgen: ah sry, didn't fully read your comment, I can add something to the documentation though

PR #241, opened 2025-01-10
chore: Migrate to pyproject.toml

tedil:
In this PR:
- the old setuptools setup.py is replaced by a pyproject.toml (e.g. for use with `uv`).
- package layout is migrated to the `src/` package layout
- github workflows for linting and testing are updated
- linting and formatting is now done via ruff (instead of flake + isort + black)
- readthedocs setup has been adjusted for v2


tedil:
Right now, 3 tests are failing:
  1. `test_archive_readme.py::test_run_archive_readme_smoke_test`: this is because the symlinks in the test directory don't point anywhere meaningful, i.e. the files simply do not exist within this repository
  2. `test_seasnap_itransfer_results.py::test_run_seasnap_itransfer_results_smoke_test`: this fails because there's some wonky logic for determining the source files from the input string (`cmd_block`) and subsequent issues with test file paths.
  3. `test_snappy_workflows.py::test_could_import_module` tests whether `snappy_pipeline` can be imported, which is not listed as a dependency anywhere.

1. can probably be fixed by ensuring these files actually exist.
2. :shrug: 
3. there's even a corresponding `test_snappy_workflows::test_could_not_import_module`; perhaps the test should be that either of `could` and `could_not` import tests succeed (and not neither nor both). And it should be mentioned somewhere that some commands are only available when there's a snappy_pipeline installation within the same environment?

Nicolai-vKuegelgen (approved):
LGTM

PR #251, opened 2025-01-29
build: pyproject.toml: Remove double biomedsheets dependency without fixed version

your-highness:
…ersion


tedil (approved):
But we should really do new releases of altamisa, biomedsheets and sodar-cli soon. I'll try to get that done this week.

Issue #256, opened 2025-02-04
Add support for cancer samplesheets


ericblanc20:
**Is your feature request related to a problem? Please describe.**
At the moment, cancer sample sheets are poorly supported in `cubi-tk` (see examples below)

**Describe the solution you'd like**
Better support for cancer projects including multi-assays (WGS, WES & mRNA_seq are frequent, possibly also including single cell data)

**Describe alternatives you've considered**
Replacing `biomedsheets` by some simple & flexible 3rd party sample description (**NOT** isa-tab, then)

**Examples of missing support for cancer samplesheets**

- `snappy/parse_sample_sheet.py` only deals with germline sample sheets (all parsing is done through family ids)
- Same with `snappy/pull_sheets.py` & with `snappy/check_local.py`.
- In `sodar/pull_raw_data.py`, the `on_visit_material` method of the `LibraryInfoCollector` class hard-codes references to family ids, which are limited to germline samplesheets.
- In `snappy/itransfer_common.py`, the `IndexLibrariesOnlyMixin` class is only implemented for germline cases
- The `snappy/itransfer_sv_calling.py` & `snappy/itransfer_variant_calling.py` are only valid for germline (Note that `snappy/itransfer_step.py` is an attempt to transfer the output of any step to SODAR. If properly extended & tested, it could replace the other `snappy/itransfer_xxx.py` files).

Nicolai-vKuegelgen:
Some comments

> #### `cubi-tk sodar pull-data-collection`
> Fails with error, although the assay UUID is correct.

This is a separate issue from the cancer samplesheets, it is defenitely a bug thoug. I'd recommend a separate ticket for that.

> #### `cubi-tk sodar pull-raw-data`
> Fails, no sample is found, but no error with assay UUID.

I actually don't know what the sodar pull-raw-data function is supposed to do, exactly. We might have to re-imagine that?

> #### `landing-zone-create`
> Works, but doesn't respect the `--dry-run` directive (landing zone is always created), and the collections for each library are not created (perhaps on purpose?)

The current old SodarAPI implementation can not automatically create collections. If the API can do that in general, I think we should change the default while switching to a newer implementation.
This is however also unrelated to cancer samplesheets.


> #### `ingest-fastq`
> Works fine (but the collection & internal file structure are not automagically setup)

Not sure which file structure you expect? If its the Sodar base collections, then see above.


> ### `cubi-tk snappy`

> #### `pull-processed-data`
[...]

Issue #164, opened 2023-04-24
`cubi-tk snappy pull-sheets` has undocumented requirements


Nicolai-vKuegelgen:
**Is your feature request related to a problem? Please describe.**

When trying to use snappy  &  `cubi-tk snappy pull-sheets` functions for projects not related to medical genetics other samplesheet templates can be more useful. However, it is not clearly documented which characteristics of the templates are actually necessary for `cubi-tk snappy pull-sheets` (and by extension snappy) to run without errors. 

By trial and error I have now discovered the following constraints
- The 'Library Name' column in Sodar needs to contain a substring like '-N1-DNA1-WGS1' that identifies the library type. This reqiurement may be exclusive to the cubi-tk call and not actually affect snappy.
- The following source 'Characteristics' columns need to be present and - at least in partially - filled with specific non-empty values:
  - Disease Status (allowed: "affected", "carrier", "unaffected", "unknown")
  - Sex (allowed: "female", "male", "unknown")
  - Mother
  - Father
  - Batch

**Describe the solution you'd like**

Ideally a more flexible template that only requires the minimum information required for snappy to run should be added.

**Describe alternatives you've considered**

At a minimum these requirements should be documented such that they can be easily looked up.

**Additional context**
Snappy currently can't handle dashes in Source names: [snappy-391](https://github.com/bihealth/snappy-pipeline/issues/391)
If this can not be fixed easily on the snappy side, `cubi-tk snappy pull-sheets` could give a warning in these cases.

sellth:
Only Sex and Disease Status are required columns as they are not filled with default values in case of a no-show: https://github.com/bihealth/cubi-tk/blob/3939f83bf510d0dfdcfd698234c27f550342c20e/cubi_tk/snappy/pull_sheets.py#L304-L318

Easy fix would be to also set `or "." / "0"` for these values.

mbenary:
The column source.affected is only relevant for germline and thus does not exist in the somatic case. And `cubi-tk snappy pull-sheets` fails in line 310.

PR #258, opened 2025-02-13
fix: [#256] cancer and multiassay support pull_sheets.py

slaemmle:
added 

- multiassay support with new "--assay-txt" arg
- cancer biomedsheet support 

to pull_sheets and expanded existing tests


Nicolai-vKuegelgen (changes requested):
Overall this looks quite good, and I don't see any issues with the code changes itself. I also particularly like moving the functions for ISA parsing out of the snappy functions itself.

There is one potential problem I see, that is not covered by the tests:
From what I understand the current ISA parsing functions read the whole ISA investigation, including the investigation file which should contain references to all assay files in a multi-assay project. Since the assay selection is now taking place by selecting only specific assay files, I am not sure if the ISA parser will actually work properly.
Therefore a test with a multi-assay investigation is also needed to make sure assay selection will work in practice.

Otherwise, we should update the documentation of this function, so that it's clear which requirements are tied to each samplesheet template (see also #164 ).
An update to our docs would be ideal, but those are quite of of date, so maybe additional description in the command line help is easier to add right now.

Nicolai-vKuegelgen (approved):
LGTM

lets try to remember adding a test with multiple assays

PR #259, opened 2025-02-13
fix: [#256] support for multiple studies and multiple assays in sodar_common.py

slaemmle:
added multi study support to assay search in sodar_common.py


Nicolai-vKuegelgen (approved):
LGTM

A project with multiple studies will (essentially) always also have multiple studies, so the added comment could be updated, but that is a very minor thing.

PR #264, opened 2025-02-17
fix: [#256] adapt check_local for cancer and pull_sheets fix

slaemmle:
Added to check_local.py:
- if tsv-shortcut == "cancer" : check, if every patient has at least one normal and one tumor sample
- Added new tests

pull-sheets.py and test: 
- fix: samplename was missing number
- assay-uuid intead of assay-txt as commandline parameter
- Multiassay test added
- warning added if more than one assay is present


src/cubi_tk/snappy/pull_sheets.py
```diff
@@ -173,7 +173,21 @@ def build_sheet(
...
+            project_uuid=project_uuid,
+        )
+        assay = None
+        for study in investigation.studies.values():
+            for remote_assay_uuid in study.assays.keys():
+                if assay_uuid== remote_assay_uuid:
+                    assay = study.assays[remote_assay_uuid]
+                    assay_filename = assay.file_name
+                    break
+    isa = isa_dict_to_isa_data(isa_dict, assay_filename)
```
Nicolai-vKuegelgen, line 190: If the given assay UUID doesn't exist/match anything, then from what I see cubi-tk will simply (and I think silently) load all assays here.
There sould be at least a warning, maybe bettern an error if the given assay_uuid doesn't match (one of multiple UUIDs)

src/cubi_tk/snappy/pull_sheets.py
```diff
@@ -137,11 +137,11 @@ def setup_argparse(parser: argparse.ArgumentParser) -> None:
...
 
     parser.add_argument(
-        "--assay-txt",
+        "--assay-uuid",
         default=None,
         required=False,
         type=str,
-        help="Assay Name e.g. 'a_project_exome_sequencing.txt' if multiple assays are present",
+        help="Assay UUID for assay if multiple assays are present",
     )
```
Nicolai-vKuegelgen, line 145: This will apply to ALL projects defined in the snappy config, however it will only make sense to one.
Maybe add a warning somewhere if multiple projects are foud while assay_uuid is defined?

src/cubi_tk/snappy/pull_sheets.py
Nicolai-vKuegelgen: Additionally (somewhat adressing the multiple project issue): if only one assay is found for agiven project/investigation, then maybe that assay should always be used, regardless of assay_uuid definition? (or maybe give a warning but ignore assay_uuid)

Nicolai-vKuegelgen (changes requested):
Overall this looks good!

There are two potential problems I can see:
- There is so far no handling of cases where the given assay-uuid does not exist/match any assay
- snappy config files may define more than one porject/investigation. If his occurs while assay uuid is defined, this also needs to be handeled/mitigated somehow (since the UUID can only work properly for one project). Restricting the assay option to single investigation snappy config files should be okay for now as a baseline solution.

Nicolai-vKuegelgen (approved):
LGTM

PR #265, opened 2025-02-17
fix: [256] adapt sodar pull-raw-data for cancer

slaemmle:
adapt sodar pull-raw-data for cancer
fix for snappy pull-all-data that relies on sodar pull-raw-data


Nicolai-vKuegelgen (approved):
LGTM

This is an important change actually, as the "Library Name" column has been removed from all ISA trmplates used by Sodar or cubi-tk (it is not actually defined by the ISA standard)

PR #276, opened 2025-03-06
refactor: improve parser

slaemmle:
Normalized parser arguments and improved --help output
Removed Config classes
fixed .git-ignore(removed /src) and fixed resulting linting errors in /src


src/cubi_tk/sodar/lz_create.py
```diff
@@ -109,10 +92,10 @@ def execute(self) -> typing.Optional[int]:
             lz = existing_lzs[-1]
         else:
             lz = api.landingzone.create(
-                sodar_url=self.args.sodar_url,
+                sodar_url=self.args.sodar_server_url,
                 sodar_api_token=self.args.sodar_api_token,
                 project_uuid=self.args.project_uuid,
-                assay_uuid=self.args.assay,
+                assay_uuid=self.args.assay_uuid,
```
Nicolai-vKuegelgen, line 98: More of a note for future refactoring: here it is assumed that api.landingzone.create can handle 'assay_uuid = None' (and presumably fail on multiple assays). I think this is something we should change in the future (i.e. make sure assay_uuid isn't None before calling the API).

src/cubi_tk/sodar/pull_raw_data.py
```diff
@@ -11,35 +11,15 @@
...
 
-from ..common import load_toml_config
+from cubi_tk.parsers import check_args_sodar_config_parser, print_args
+
 from ..isa_support import (
     InvestigationTraversal,
     IsaNodeVisitor,
     first_value,
     isa_dict_to_isa_data,
 )
```
Nicolai-vKuegelgen, line 21: Refactoring note: why does this need ISA traversal?

src/cubi_tk/sodar_api.py
```diff
@@ -1,47 +1,32 @@
...
+
+from .common import is_uuid
 from .exceptions import ParameterException, SodarAPIException
 
 
 class SodarAPI:
-    def __init__(self, sodar_url: str, sodar_api_token: str, project_uuid: str):
-        self.sodar_url = sodar_url
+    def __init__(self, sodar_server_url: str, sodar_api_token: str, project_uuid: str, config = None):
+        self.sodar_server_url = sodar_server_url
```
Nicolai-vKuegelgen, line 17: maybe there's an argument to made here, to just pass the "args" object (which now should always use the same structure for sodar related arguments)

src/cubi_tk/sodar_api.py
```diff
@@ -1,47 +1,32 @@
...
-                    "SODAR API token not found in config files. Please specify on command line."
+        any_error, args= check_args_sodar_config_parser(argparse.Namespace(config=config, sodar_server_url=self.sodar_server_url, sodar_api_token= self.sodar_api_token))
+        self.sodar_server_url = args.sodar_server_url
+        self.sodar_api_token = args.sodar_api_token
+        if any_error:
+            raise ParameterException(
+                    "SODAR variables not found in config files. Please specify on command line."
                 )
         if not is_uuid(self.project_uuid):
             raise ParameterException("Sodar Project UUID is not a valid UUID.")
```
Nicolai-vKuegelgen, line 32: I think this could just be moved to the `check_args_sodar_config_parser`, it should be either generally applicable or is probably not needed anymore.
While, the project_uuid part is not yet included in the "predefeined" sodar parser, it is however something the API will _always_ need and I don't think the sodar url and token can be used without a project uuid.
(We will also need some special handling for parsers that defined DESTINATIOn instead of project uuid, but I think a generalised check_args_sodar_config_parser (see also my comment there) can handle that.

tests/test_snappy_pull_sheets.py
```diff
@@ -64,57 +63,78 @@ def return_api_investigation_mock():
...
+            last_batch= None,
+            tsv_shortcut= "cancer",
+            assay_uuid= None
+    )
     path = pathlib.Path(__file__).resolve().parent / "data" / "pull_sheets" / "sheet_cancer.tsv"
     with open(path, "r") as file:
         expected = "".join(file.readlines())
     mocker.patch(
         "sodar_cli.api.samplesheet.export", return_value=load_isa_dict("isa_dict_cancer.txt")
     )
```
Nicolai-vKuegelgen, line 114: Why not use the `@patch` decorator here? (from unittest.mock import patch)
I guess either way is fine. Some of the (newer) tests inc ubi-tk use the decorator approach, the older ones use various mocker approaches (often without using the mocker fixture though), so we should decide on one approach to go with.

Nicolai-vKuegelgen (commented):
Very nice work overall!

Most of my comments are either notes regarding future/additional refactoring steps or discussion on how to integrate/improve the current changes with what will/might come next.

Apart from the few specific things that could directly be improved, I think we can merge this into a (new?) dev branch and start working on more refactoring from there.
Not sure how to best keep track of all the things, I now noticed will need refactoring, is there a good solution to keep track of that?

src/cubi_tk/parsers.py
```diff
@@ -0,0 +1,162 @@
...
+def print_args(args: argparse.Namespace):
+    token = args.sodar_api_token
+    args.sodar_api_token = "****"
+    logger.info("Args: {}", args)
+    args.sodar_api_token = token
+
+
+basic_config_parser = argparse.ArgumentParser(description="The basic config parser", add_help=False)
+basic_group = basic_config_parser.add_argument_group("Logging Configuration")
+basic_group.add_argument("--verbose", action="store_true", default=False, help="Increase verbosity.")
```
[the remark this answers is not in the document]
slaemmle: As fa as i understand the code the parsers get constructed in main and so for every cubi-tk execution all the different parsers will be created anyways
I wrote it like this to not have e.g. the basic parser get created 5 times in sodar, snappy, etc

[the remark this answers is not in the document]
Nicolai-vKuegelgen: ah true, that makes sense

src/cubi_tk/sea_snap/pull_isa.py
```diff
@@ -9,33 +9,22 @@
...
 import typing
 from uuid import UUID
 
 from loguru import logger
 import requests
 
-URL_TPL = "%(sodar_url)s/samplesheets/api/remote/get/%(project_uuid)s/%(api_key)s?isa=1"
+from cubi_tk.parsers import check_args_sodar_config_parser, print_args
+
+URL_TPL = "%(sodar_server_url)s/samplesheets/api/remote/get/%(project_uuid)s/%(api_key)s?isa=1"
```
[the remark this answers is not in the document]
slaemmle: Will add TODOs in the Code for stuff like this

src/cubi_tk/snappy/pull_sheets.py
```diff
@@ -148,7 +98,7 @@ def setup_argparse(parser: argparse.ArgumentParser) -> None:
 def check_args(args) -> int:
     """Argument checks that can be checked at program startup but that cannot be sensibly checked with ``argparse``."""
     any_error = False
-
+    any_error, args =  check_args_sodar_config_parser(args)
     # Postprocess arguments.
     if args.library_types:
         args.library_types = args.library_types.split(",")  # pragma: nocover
```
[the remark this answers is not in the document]
slaemmle: If i understand the code right it is working like a filter
So if you specify nothing yourself, everything gets pulled( same as if you specify every library type) and if you specify a libtype only sample with this lib type get pulled

currently the default is just an empty list, which acts the same as if it was a list with all possible lib types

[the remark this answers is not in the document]
Nicolai-vKuegelgen: mmh, I would rather make the default an explicitly filled list then. Currently it's very much not intuitive what the default behavior (or value) here actually is.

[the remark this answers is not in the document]
slaemmle: yes, i changed it in the parser to default as empty list without extra processing afterwards

slaemmle:
Additional Tasks before merging into any kind of production Branch:
- [ ] update Documentation with commandline args
- [ ] run systemtests (integrationtests might not fully cover all the changes)

src/cubi_tk/snappy/pull_sheets.py
```diff
@@ -63,7 +65,9 @@ def setup_argparse(parser: argparse.ArgumentParser) -> None:
     )
 
     parser.add_argument(
-        "--library-types", choices=("WES", "WGS", "Panel_seq"), help="Library type(s) to use, comma-separated, default is to use all."
+        "--library-types",
+        nargs= "*",  # same as default = []
+        help="Library type(s) to use( e.g. WES, WGS, or Panel_seq) paassed like '--library-types WES Panel_seq', default is to use all."
```
Nicolai-vKuegelgen, line 70: This should keep the `choices` argument, handing _any_ kind of string to the isa downloader is going to cause problems otherwise

tests/test_sodar_ingest_fastq.py
```diff
@@ -203,8 +203,9 @@ def test_run_sodar_ingest_fastq_smoke_test(mocker, requests_mock, fs):
         "--sodar-api-token",
         "XXXX",
         "--yes",
-        fake_base_path,
         landing_zone_uuid,
+        fake_base_path,
+
```
Nicolai-vKuegelgen, line 208: Does this mean the order of positional arguments has now changed? Can we avoid that somehow?
(Applies to all commands where this is the case)

tests/test_sodar_ingest.py
```diff
@@ -51,7 +51,7 @@ def ingest(fs):
         "target",
     ]
 
-    sodar_parser = get_sodar_parser()
+    sodar_parser =  get_sodar_parser(with_dest= True, dest_string="destination", help_string="UUID from Landing Zone or Project - where files will be moved to.")
```
Nicolai-vKuegelgen, line 54: extra space here?

Nicolai-vKuegelgen (commented):
Looks very good overall.

Is there any possibility we can keep the current order of all positional arguments or is this not possible when handing through the project UUID  argument?

tests/test_sodar_ingest_fastq.py
slaemmle: Yes, the order has changed, which is a bit of annoying but i haven't found an easy way to avoid that..

Nicolai-vKuegelgen: okay let's leave it like this then and merge into the dev branch.
We should start making a change log of cases this like, so we have a full list of "breaking changes" (if it's just adjustment of how cubi-tk needs to be called)

PR #277, opened 2025-03-14
Refactor: assay and study interactive

slaemmle:
Assay-uuid added to global argparse

When mutliple studies and assays are present, the user can select which one to choose (if not --yes)

### DIFF
--- a/src/cubi_tk/sodar_api.py
+++ b/src/cubi_tk/sodar_api.py
@@ -26,21 +26,21 @@ def get_assay_from_uuid(sodar_server_url, sodar_api_token, project_uuid, assay_u
                 logger.info(f"Using provided Assay UUID: {assay_uuid}")
                 assay = study.assays[assay_uuid]
                 return assay
-            #will only iterate through first study, if multiple studys present
-            assays_ = list(study.assays.keys())
-            #only one assay
-            if len(assays_) == 1:
-                return study.assays[assays_[0]]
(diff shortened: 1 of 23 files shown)


src/cubi_tk/sodar_api.py
```diff
@@ -13,23 +13,99 @@
...
+    logger.warning("No --assay-uuid specified but multiple assays present, which assay do you want to choose?")
+    i = 0
+    while i < len(assay_uuids):
+        logger.warning("{}: {}", i+1, assay_uuids[i] )
+        i+=1
+    assay_num = 0
+    while (assay_num<= 0 or assay_num> len(assay_uuids)):
+        assay_num = input("Please enter the index of the Assay UUID (e.g 2):")
+        assay_num =int(assay_num)
+    return assay_uuids[assay_num-1]
```
Nicolai-vKuegelgen, line 95: Is there an option to display both UUID and assay/study names? Mnually comparing UUIDs against sodar will be cumbersome for at least some people

Nicolai-vKuegelgen (approved):
LGTM

If possible I'd add Assay & study names for user selection, but we can stay with UUIDs for now

src/cubi_tk/sodar_api.py
slaemmle: Good idea, i added a todo and will handle this in the upcoming sodarapi PR

PR #283, opened 2025-04-30
feat: major cubi-tk refactoring

slaemmle:
major refactoring includes:

- improved parser (unified commands and arguments)
- assay and study selection interactive, when multiple present
- irods refactoring
- sodar_cli replaced with own sodarapi
- cubi_tk oncology workflow refactored
- ingestdata more flexible with column selection ( columns from study now possible)
- renaming/deprecating of some cubi-tk commands
- multile sodar instances in toml possible
- hshscheme read from irods_environment.json (not fixed to.md5 anymore)

### DIFF
--- a/tests/test_snappy_itransfer_variant_calling.py
+++ b/tests/test_snappy_itransfer_variant_calling.py
@@ -146,7 +146,7 @@ def test_run_snappy_itransfer_variant_calling_smoke_test(
     mock_value = mock.MagicMock()
     mocker.patch("cubi_tk.common.Value", mock_value)
 
-    mocker.patch("cubi_tk.sodar.ingest_data.iRODSTransfer.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
+    mocker.patch("cubi_tk.snappy.itransfer_common.iRODSCommon.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
 
     # Actually exercise code and perform test.
     parser, _subparsers = setup_argparse()
--- a/tests/test_snappy_itransfer_ngs_mapping.py
+++ b/tests/test_snappy_itransfer_ngs_mapping.py
@@ -139,7 +139,7 @@ def test_run_snappy_itransfer_ngs_mapping_smoke_test(
 
     mock_check_call = mock.mock_open()
     mocker.patch("cubi_tk.common.check_call", mock_check_call)
-    mocker.patch("cubi_tk.sodar.ingest_data.iRODSTransfer.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
+    mocker.patch("cubi_tk.snappy.itransfer_common.iRODSCommon.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
 
     # Actually exercise code and perform test.
     res = main(argv)
--- a/tests/test_snappy_itransfer_raw_data.py
+++ b/tests/test_snappy_itransfer_raw_data.py
@@ -121,7 +121,7 @@ def test_run_snappy_itransfer_raw_data_smoke_test(
     mocker.patch("cubi_tk.snappy.common.open", fake_open)
     mocker.patch("cubi_tk.common.open", fake_open)
 
-    mocker.patch("cubi_tk.sodar.ingest_data.iRODSTransfer.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
+    mocker.patch("cubi_tk.snappy.itransfer_common.iRODSCommon.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
 
 
     # Actually exercise code and perform test.
--- a/tests/test_snappy_itransfer_step.py
+++ b/tests/test_snappy_itransfer_step.py
@@ -155,7 +155,7 @@ def test_run_snappy_itransfer_step_smoke_test(
     mock_check_call = mock.mock_open()
     mocker.patch("cubi_tk.common.check_call", mock_check_call)
 
-    mocker.patch("cubi_tk.sodar.ingest_data.iRODSTransfer.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
+    mocker.patch("cubi_tk.snappy.itransfer_common.iRODSCommon.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
 
     # Actually exercise code and perform test.
     res = main(argv)
--- a/tests/test_snappy_itransfer_sv_calling.py
+++ b/tests/test_snappy_itransfer_sv_calling.py
@@ -259,7 +259,7 @@ def test_run_snappy_itransfer_sv_calling_smoke_test(mock_transfer, mocker, germl
     mock_check_call = mock.mock_open()
     mocker.patch("cubi_tk.common.check_call", mock_check_call)
 
-    mocker.patch("cubi_tk.sodar.ingest_data.iRODSTransfer.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
+    mocker.patch("cubi_tk.snappy.itransfer_common.iRODSCommon.irods_hash_scheme", mock.MagicMock(return_value="MD5"))
 
     # Actually exercise code and perform test.
     parser, _subparsers = setup_argparse()


src/cubi_tk/common.py
```diff
@@ -1,80 +1,124 @@
...
     if verbose:
-        logger.info("Computing md5 hash for {}".format(filename))
+        logger.info(f"Computing {hash_scheme} hash for {filename}")
     the_hash = None
     with open(filename, "rb") as f:
-        the_hash = hashlib.md5()
+        if hash_scheme.lower() == "md5":
+            the_hash = hashlib.md5()
+        else: #currently only md5 and SHA256 supported
+            the_hash= hashlib.sha256()
```
Nicolai-vKuegelgen, line 46: we should still be explicit with sha256 here, just to make this future proof

src/cubi_tk/sodar/check_remote.py
```diff
@@ -1,10 +1,10 @@
 """``cubi-tk sodar check-remote``: check that files are present in remote SODAR/iRODS.
 
-This tool searches recursively for all files with associated .md5 files in a given directory (or PWD)
-and compares them against files linked to a given SODAR project/irods directory, also the md5 sums
-recorded in local .md5 files can be rechecked. No hash format other than md5 is currently supported.
-Comparison is done based (first) on file names and then (unless disabled) on md5 sums,
-therefore wiles with different names but the same md5 sum will not be matched.
+This tool searches recursively for all files with associated checksums files in a given directory (or PWD)
+and compares them against files linked to a given SODAR project/irods directory, also the checksums
+recorded in local checksum files can be rechecked. No hash format other than md5 and sha256is currently supported.
```
Nicolai-vKuegelgen, line 5: misssing space

src/cubi_tk/sodar/ingest_data.py
```diff
@@ -371,90 +265,124 @@ def check_col_index(column_index):
...
+    def find_collection_name(self, sample_name, column_match, m):
+        if column_match is None:
+            return sample_name
+        val = column_match[sample_name]
+        if not isinstance(val, list):
+            return val
+        #needs further matching with tumor and conservation method
+        conservation_col_name = None
+        conservation = m.groupdict(default=None)["conservation"]
+        tissue = m.groupdict(default=None)["tissue"]
```
Nicolai-vKuegelgen, line 347: Maybe add a check here if these capture groups exist?
Probably catched later anyway, but might be good to raise an explicit message what went wrong?

src/cubi_tk/sodar/ingest_data.py
```diff
@@ -468,15 +396,11 @@ def build_jobs(self, library_names=None) -> tuple[str, tuple[TransferJob, ...]]:
                 real_path = os.path.realpath(path)
                 if not os.path.isfile(real_path):
                     continue  # skip if did not resolve to file
-                if real_path.endswith(".md5"):
+                if real_path.endswith((".md5", ".md5sum", ".sha256")):
```
Nicolai-vKuegelgen, line 399: do we generally treat '.md5sum' the same as '.md5'? 
Also, i think this will prevent us from uploading md5 files into an irods that uses sha256 (but maybe we want to archive the md5 anyway).

I would ONLY skip the proper hash scheme here (since these are the files that are added automatically a bit below), all other files should not be excluded from Sodar upload

src/cubi_tk/sodar/ingest_data.py
```diff
@@ -509,15 +432,15 @@ def build_jobs(self, library_names=None) -> tuple[str, tuple[TransferJob, ...]]:
...
                         )
                         logger.error(msg)
-                        raise ParameterException(msg)
+                        raise ParameterException(msg) from KeyError
 
                     # This was the original code, but there is no need to change the remote file names once they are
                     # mapped to the correct collections:
                     # remote_file = str(remote_file)
                     # for m_pat, r_pat in self.args.remote_dir_mapping:
                     #     remote_file = re.sub(m_pat, r_pat, remote_file)
```
Nicolai-vKuegelgen, line 442: We can remove this now

src/cubi_tk/sodar/ingest_data.py
```diff
@@ -534,9 +457,10 @@ def execute(self) -> typing.Optional[int]:
             return res
 
         logger.info("Starting cubi-tk sodar {}", self.command_name)
-        logger.info("args: {}", self.args)
-
-        lz_uuid, transfer_jobs = self.build_jobs()
+        print_args(self.args)
+        irods_hash_scheme = iRODSCommon(sodar_profile=self.args.config_profile).irods_hash_scheme()
```
Nicolai-vKuegelgen, line 461: not 100% happy about having creating a new instance here.
Maybe consider adding the option to instantiate iRODSTransfer before adding all the jobs to it?
(don't do this now, maybe a make an issue?)

src/cubi_tk/sodar_api.py
```diff
@@ -1,81 +1,109 @@
...
+
+    :param assays: Assays UUIDs as found in Studies.
+    :type assays: list
+    """
+    logger.warning("Multiple studies present, which one do you want to choose?")
+    i = 0
+    while i < len(study_uuids):
+        study_name = studies[study_uuids[i]].file_name
+        logger.warning("{}: {}", i+1, study_name)
+        i+=1
```
Nicolai-vKuegelgen, line 34: this could be a for loop:
for i, study_uuid in enumerate(study_uuids)

src/cubi_tk/sodar_api.py
```diff
@@ -1,81 +1,109 @@
...
+
+    :param assays: Assays UUIDs as found in Studies.
+    :type assays: list
+    """
+    logger.warning("No --assay-uuid specified but multiple assays present, which assay do you want to choose?")
+    i = 0
+    while i < len(assay_uuids):
+        assay_name = assays[assay_uuids[i]].file_name
+        logger.warning("{}: {}", i+1, assay_name)
+        i+=1
```
Nicolai-vKuegelgen, line 54: for loop as above

src/cubi_tk/sodar_api.py
```diff
@@ -1,81 +1,109 @@
...
+    study_num = 0
+    while (study_num<= 0 or study_num> len(study_uuids)):
+        study_num = input("Please enter the index of the Study (e.g 2):")
+        study_num =int(study_num)
+    study_uuid = study_uuids[study_num-1]
+    logger.info("Chosen Study: {}", studies[study_uuid].file_name)
+    return study_uuid
+
+def get_user_input_assay_uuid(assay_uuids:list[str], assays: dict[str, api_models.Assay]) -> UUID:
```
Nicolai-vKuegelgen, line 43: This and the study function could probably be refactored into one

tests/test_snappy_itransfer_variant_calling.py
```diff
@@ -130,22 +132,30 @@ def test_run_snappy_itransfer_variant_calling_smoke_test(
...
     # Actually exercise code and perform test.
     parser, _subparsers = setup_argparse()
     args = parser.parse_args(argv)
     res = main(argv)
     assert not res
-    mock_transfer.assert_called_with(expected_tfj, ask=not args.yes)
+    mock_transfer.assert_called_with(expected_tfj, ask=not args.yes, sodar_profile='global')
     mock_transfer_obj.put.assert_called_with(recursive=True, sync=args.overwrite_remote)
-
+    print(fs)
```
Nicolai-vKuegelgen, line 158: should probably be taken out

src/cubi_tk/sodar/check_remote.py
```diff
@@ -56,36 +59,36 @@ def run(self):
...
                 continue
 
-            with open(md5file, "r", encoding="utf8") as f:
-                md5sum = f.readline()
+            with open(checksumfile, "r", encoding="utf8") as f:
+                checksum = f.readline()
                 # Expected format example:
                 # `459db8f7cb0d3a23a38fdc98286a9a9b  out.vcf.gz`
-                md5sum = md5sum.split(" ")[0]
+                checksum = checksum.split(" ")[0]
```
Nicolai-vKuegelgen, line 79: maybe add a check that this fits (I think the hash schemes also define expected regex/pattern?)

Nicolai-vKuegelgen (changes requested):
Mostly everything okay (since we already reviewed it anyway)

A few things with the checksum changes, otherwise minor things I may have just seen now.

Maybe check with Eric regarding the failed linting - looks like a typo in the variable name?

src/cubi_tk/sodar/ingest_data.py
slaemmle: the dragen analysis generates .md5sum
Im not sure if we could use those generated ones instead of .md5
this is to prevent something like generation of .md5sum.md5 or .md5sum.sha256
but i will change it to the used hash_scheme + .md5sum

Nicolai-vKuegelgen: ah yeah that's a good point. Please add this explanation also to the code as a comment (&/ make an additional variable for file endings to generally ignore for transfers, we could even make a cmd-line option to add to that at some point)

src/cubi_tk/sodar/check_remote.py
slaemmle: the hash itself is longer in sha256 vs md5, but there will still  be a space before the file ending, which is what we are looking for here

Nicolai-vKuegelgen: yeah I mostly mean to check the expected length, those are already defined somewhere in cubi-tk IIRC

Nicolai-vKuegelgen (approved):
LGTM

Thanks this looks very good!

PR #284, opened 2025-05-06
fix: switch sodar profile with different irods_env

slaemmle:
added last_env.json to check what the last authenticated sodar/irods instance was

### DIFF
--- a/src/cubi_tk/irods_common.py
+++ b/src/cubi_tk/irods_common.py
@@ -84,12 +84,13 @@ class iRODSCommon:
 
     def _init_irods(self) -> iRODSSession:
         """Connect to iRODS. Login if needed."""
-        self._check_and_gen_irods_files()
+
         count_tries = 1
         while True:
             try:
                 session = iRODSSession(irods_env_file=self.irods_env_path)
                 session.connection_timeout = 600
+                self._check_and_gen_irods_files()
                 return session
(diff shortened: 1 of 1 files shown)


src/cubi_tk/irods_common.py
```diff
@@ -106,13 +107,22 @@ def _check_and_gen_irods_files(self, overwrite = False):
...
             #check if irodsfile exists
             irodsA_path = self.irods_env_path.parent.joinpath(".irodsA")
+            last_profile_path = self.irods_env_path.parent.joinpath("last_profile.json")
             if (irodsA_path.exists()):
                 self.irodsA_file_found = True
+                if last_profile_path.exists():
+                    with open(last_profile_path) as last_profile_file:
+                        last_used_env = json.load(last_profile_file)["last_used_env"]
+                        overwrite = last_used_env != str(self.irods_env_path) #overwrite irodsA file if last authenticated profile is different to current
+                elif not str(self.irods_env_path).endswith("irods_environment.json"):
```
Nicolai-vKuegelgen, line 117: I think this could also be something like `self.irods_env_path.basename != "irods_environment.json"`

Nicolai-vKuegelgen (approved):
Overall LGTM

I took me a moment to understand that "last_profile.json" saves the name of the last profile and not the profile itself, maybe add a comment to make that clearer?

PR #285, opened 2025-05-09
fix: get assay UUID pull data

slaemmle:
Bugfix:

- RetrieveSodarCollection was not initialised right
- Assay UUID was queried in RetrieveSodarCollection, but never set


mbenary (approved):
LGTM.

PR #289, opened 2025-06-25
feat: changes for onko data upload

slaemmle:
Changed regex and column matching to data structure of onko files

### DIFF
--- a/src/cubi_tk/sodar/ingest_data.py
+++ b/src/cubi_tk/sodar/ingest_data.py
@@ -341,11 +341,14 @@ class SodarIngestData(SnappyItransferCommandBase):
         logger.debug(val)
         if not isinstance(val, list):
             return val
+        if not self.args.preset == "onk_analysis" :
+            logger.error("preset onk_analysis needs to be set, please check your input parameters")
+            raise KeyError
         #needs further matching with tumor
         tissue = m.groupdict(default=None)["tissue"]
         for col_col_name, col_sample_name in val:
             col_tissue = col_sample_name.split("-")[-1][0]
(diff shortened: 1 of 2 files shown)


src/cubi_tk/sodar/ingest_data.py
```diff
@@ -338,32 +337,17 @@ def _check_col_index_and_get_val(self, column_index, sheet_type, sheet_header, s
...
-        val = column_match[sample_name]
+        val = column_match[str(sample_name)]
+        logger.debug(val)
         if not isinstance(val, list):
             return val
-        #needs further matching with conservation method and if existent tumor
-        conservation_col_name = None
-        conservation = m.groupdict(default=None)["conservation"]
+        #needs further matching with tumor
         tissue = m.groupdict(default=None)["tissue"]
```
Nicolai-vKuegelgen, line 345: Please add a comment to indicate when the initial return is supposed to happen and/or just explicitly return/throw an error before the onko specific tumor checks happen.

src/cubi_tk/sodar/ingest_data.py
```diff
@@ -431,6 +415,9 @@ def build_jobs(self, hash_ending) -> tuple[str, tuple[TransferJob, ...]]:  # noq
                             collection_name=collection_name,
                             **match_wildcards,
                         )
+                        #if onko and germline analysisdata change analysis to germline_analysis
+                        if self.args.preset == "onk_analysis" and "DragenGermline" in path:
+                            remote_file.replace("analysis", "germline_analysis")
```
Nicolai-vKuegelgen, line 420: Maybe instead of hard coding this, we should consider adding an option to do string/regex replacement in the remote filenames/paths (analogous to the current -m which acts on sample name only). Then we could add "inbuilt" defaults for these to the presets. 
Doesn't have to be done right away though I guess

PR #301, opened 2025-09-16
feat: update ingest-data for wgs germline

slaemmle:


Nicolai-vKuegelgen (approved):
LGTM

This behavior should maybe also get a testcase?

PR #336, opened 2026-06-15
feat: no login required when profile switch

### LINKED ISSUE #334: Switch Sodar instance without password reentry
When switiching Sodar instances, password reentry ist needed

enable switching without passwort reentry through multiple .irodsA files and adding it to irods_env json
(https://irods.wur.nl/userguide/irods_multizone/)

slaemmle:


src/cubi_tk/irods_common.py
```diff
@@ -124,43 +127,54 @@ def _check_and_gen_irods_files(self, overwrite=False):
...
-            # check if irodsfile exists
-            irodsA_path = self.irods_env_path.parent.joinpath(".irodsA")
-            ##check path of last authorized irods_environment.json
+            # read hashscheme and irodsA path from irods env file
+            irods_env_json = {}
+            with open(self.irods_env_path) as irods_env_data:
+                irods_env_json = json.load(irods_env_data)
+                self.hash_scheme = irods_env_json["irods_default_hash_scheme"]
+                if self.hash_scheme not in HASH_SCHEMES:
+                    logger.error("Hashscheme currently not supported")
```
Nicolai-vKuegelgen: f"Hashscheme '{self.hash_scheme}' currently not supported"
raise ValueError

src/cubi_tk/irods_common.py
```diff
@@ -124,43 +127,54 @@ def _check_and_gen_irods_files(self, overwrite=False):
...
             last_profile_path = self.irods_env_path.parent.joinpath("last_profile.json")
-            if irodsA_path.exists():
+            last_profile_json = {}
+            default_irodsA = self.irods_env_path.parent.joinpath(".irodsA")
+            irodsA_path = self.irods_env_path.parent.joinpath(".irodsA_" + self.sodar_profile)
+            if os.path.exists(irodsA_path):
                 self.irodsA_file_found = True
+                # copy irodsA_<profile> to .irodsA for use by irods client
+                shutil.copy(irodsA_path, default_irodsA)
+                # check if relogin is needed
```
Nicolai-vKuegelgen, line 163: Would it work to back this up, and restore it on exit? (IIRC `__del__` for the class, or `__exit__` if we refatcor it into a context manager)
Only worth adding if the old/previous login still works after copying the file back and forth, but if it does we should add it? (This should be easy to test just with icommands & cp/mv on the cubi-HPC)

src/cubi_tk/irods_common.py
```diff
@@ -124,43 +127,54 @@ def _check_and_gen_irods_files(self, overwrite=False):
...
-                        last_used_env = json.load(last_profile_file)["last_used_env"]
-                        overwrite = (
-                            last_used_env != str(self.irods_env_path)
-                        )  # overwrite irodsA file if last authenticated profile is different to current
+                        last_profile_json = json.load(last_profile_file)
+                        timestamp = last_profile_json.get(self.sodar_profile, 0)
+                        timestamp_today = int(time())
+                        overwrite = timestamp != 0 and timestamp < (
+                            timestamp_today - 14 * 24 * 3600
+                        )
```
Nicolai-vKuegelgen, line 171: Taking the timestamp from `last_profile_json` is probably fine in most cases, but ignores a possibly more recent .irodsA_profile file from external usage - not sure if it's worth tracking that though.

Nicolai-vKuegelgen (commented):
Looks good overall.

Only point would be to check & consider if it's worth backing up and restoring the .irodsA file, to whatever it was before running cubi-tk changed it via --config-profile.

src/cubi_tk/irods_common.py
```diff
@@ -94,6 +96,19 @@ def __init__(
...
+        if self.irodsA_file_found:
+            # Remove the .irodsA file used for this profile and restore backup if exists, to avoid issues with other profiles or global config
+            default_irodsA = self.irods_env_path.parent.joinpath(".irodsA")
+            if default_irodsA.exists():
+                default_irodsA.unlink()
+            # restore backup if exists
+            irodsA_backup = self.irods_env_path.parent.joinpath(".irodsA_backup")
+            if irodsA_backup.exists():
+                shutil.copy(irodsA_backup, default_irodsA)
+                irodsA_backup.unlink()
```
Nicolai-vKuegelgen, line 112: This files are created with shuitl.copy (or even possibly beforehand by the user), so I don't think we can assume they are symlinks?

Nicolai-vKuegelgen (commented):
I'm not sure using unlink to restore .irodsA files will work.

Also It'd be nice to build a test for the irodsA copying & restoring (using fakefs this shouldn't bee a problem? The test doesn't need to cover any actual irods session creation // can mock it away)

src/cubi_tk/irods_common.py
slaemmle: documentation says "Remove this file or link"
seems to work

tests/test_irods_common.py
```diff
@@ -35,27 +36,64 @@ def test_get_irods_error(mocksession):
...
     icommon._check_and_gen_irods_files()
+
     mockpass.assert_called()
-    mocksession.assert_any_call(irods_env_file=ANY, password=password)
-    assert icommon.irods_env_path.parent.joinpath(".irodsA").exists()
+    mock_write_pam.assert_called_once()
+    assert irods_a_path.exists()
+    assert Path(Path(irods_env_file).parent / ".irodsA_global").exists()
+    assert Path(Path(irods_env_file).parent / ".irodsA_backup").exists()
+    icommon.__del__()  # to trigger cleanup of irodsA file
```
Nicolai-vKuegelgen:
```suggestion
    del icommon # to trigger cleanup of irodsA file
```

Nicolai-vKuegelgen: Deleting the object should be better than directly invoking __del__().
In a perfect world we could also test this on the shell directly (i.e. after invoking a `cubi-tk ...` command), but that isn't really doable from within pytest ...

Nicolai-vKuegelgen (approved):
LGTM

minor change on test code, but it does the same thing

PR #337, opened 2026-06-18
refactor: sodar update-samplesheet functions for easier external usage

Nicolai-vKuegelgen:


src/cubi_tk/sodar/update_samplesheet.py
```diff
@@ -29,6 +29,15 @@
 #   may include numerical suffixes for duplicate column names (e.g. "Extract Name.1")
 
 
+class IsaDataBlock(TypedDict):
+    i_path: str
+    investigation: str
+    study_key: str
+    study: pd.DataFrame
+    assay_key: str
+    assay: pd.DataFrame
```
slaemmle, line 39: To support multi study/assay projects this could also be dictionaries{ study_key: stud_df, ...}

src/cubi_tk/sodar/update_samplesheet.py
```diff
@@ -241,41 +251,44 @@ def setup_argparse(cls, parser: argparse.ArgumentParser) -> None:
...
+        """Get samplehseet from SODAR API and load as DataFrame"""
 
-        # Get samplehseet from SODAR API
-        sodar_api = SodarApi(self.args, with_dest=True)
-        print_args(self.args)
-        isa_data = sodar_api.get_samplesheet_export()
+        isa_data = self.sodar_api.get_samplesheet_export()
+        investigation_path = isa_data["investigation"]["path"]
         investigation = isa_data["investigation"]["tsv"]
         study_key = list(isa_data["studies"].keys())[0]
```
slaemmle, line 266: hardcoding with first study/assay would then also not be needed -> more flexible

src/cubi_tk/sodar/update_samplesheet.py
```diff
@@ -305,13 +318,34 @@ def execute(self) -> Optional[int]:
         )
 
         files_dict = {
-            "file_investigation": (isa_data["investigation"]["path"], investigation),
-            "file_study": (study_key, study_tsv),
-            "file_assay": (assay_key, assay_tsv),
+            "file_investigation": (isa_data_block["i_path"], isa_data_block["investigation"]),
+            "file_study": (isa_data_block["study_key"], study_tsv),
```
slaemmle: adapt for multi assay projects (possibly study too)

slaemmle (changes requested):

src/cubi_tk/sodar/update_samplesheet.py
Nicolai-vKuegelgen: oh right, this would currently remove all non-selected studies / assays. I'll fix that

src/cubi_tk/sodar/update_samplesheet.py
Nicolai-vKuegelgen: The `IsaDataBlock` is only a different representation of the output of `sodar_api.get_samplesheet_export(get_all=False)`, which is always only one study and one assay (and uses the shared functionality for assay selection to get there). 
The `update-samplesheet` fucntion itself does not (and probably should not) allow modifying more than one study & assay table at a time (since that really won't work without a full representation of the isa graph).

src/cubi_tk/sodar/update_samplesheet.py
Nicolai-vKuegelgen: see above

src/cubi_tk/sodar/update_samplesheet.py
slaemmle: Not sure if it would remove them or if the POST request would fail, but we need t support mulitassay here
thanks!

slaemmle (approved):
Looks good now, thank you
Might still be good to keep the possibility of editing more than one assay in mind for the future to support e.g. mixed runs with data for both assays

Nicolai-vKuegelgen:
> Might still be good to keep the possibility of editing more than one assay in mind for the future to support e.g. mixed runs with data for both assays

In principle I agree. However, this function currently does updates by matching column names to either study or assay table (whereever they can be found), and adds new samples to both. Doing the same thing for 2 assays at the same time would require a major if not complete rewrite of the function.
I think rigth now the easiest approach would be to do separate updates for each assay

PR #343, opened 2026-07-15
fix: solve `sodar update-samplesheet` issues and add improvements

### LINKED ISSUE #268: `cubi-tk sodar update samplesheet` autofill does not work correctly on small/template projects
**Describe the bug**
`cubi-tk sodar update samplesheet` should autofill values of new additions to a smaplehseet, if those values are unique in the existing sheet.
In a very small project (i.e. created from a template) this is the case for almost all values, and behaviour is not as intended:
- autofill takes precedence over sample specific metadata (i.e. affected status)
- autofill takes precedence over specifically supplied static metadata columns.
Additionally autofill should maybe also work for the columns defining ontology term IDs (if not data was added to those in a given set and they were unique before)

**To Reproduce**
1. Create a template project
2. Add new samples with `cubi-tk sodar update samplesheet` 
4. Incorrect metadata will be autofilled

**Expected behavior**
Autofill should never overwrite user given data

Nicolai-vKuegelgen:
- Fixes precedence of column auto-filling over actual values given
- adds dryrun option

closes #268

### DIFF
--- a/src/cubi_tk/common.py
+++ b/src/cubi_tk/common.py
@@ -329,7 +329,7 @@ def print_line(line):
 
 
 def _overwrite_helper_show_diff(
-    lines, new_lines, out_file, out_path, out_path_obj, show_diff_side_by_side, old_lines = None
+    lines, new_lines, out_file, out_path, out_path_obj, show_diff_side_by_side, old_lines=None
 ):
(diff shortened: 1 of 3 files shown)


Nicolai-vKuegelgen:
Note: test for dryrun is somewhat incomplete.

slaemmle (approved):

Issue #341, opened 2026-07-07
Bug report: `multi_assay_study_warning()` called with invalid `assays=` keyword → `TypeError`


steinbrecht:
**Component:** cubi-tk (`sodar_api.py`, `get_assay_from_uuid()`)
**Version:** cubi-tk 0.7.0 (commit `d4d32ba`)

## Summary

In the "no `--assay-uuid` given, `--yes`, multiple assays" auto-selection branch
of `SodarApi.get_assay_from_uuid()`, the helper `multi_assay_study_warning` is
called with a keyword argument (`assays=`) that does not exist in its signature,
raising a `TypeError` and aborting the command.

## Details

`multi_assay_study_warning` is defined as:

```python
def multi_assay_study_warning(content: dict, string: str = "Assays") -> None:
    ...
```

Its first parameter is `content`; there is no `assays` parameter. The call site
in `get_assay_from_uuid()` passes `assays=assays_`:

```python
else:
    # no assay specified, return first one or ask user
    assays_ = list(study.assays.keys())
    if len(assays_) == 1 or self.yes:
        assay = study.assays[assays_[0]]
        self.assay_uuid = assays_[0]
        if self.yes and len(assays_) > 1:
            multi_assay_study_warning(assays=assays_)   # <-- invalid keyword
        return assay, study
```

The other call site is correct and passes positionally:
`multi_assay_study_warning(study_keys, string="studies")`.

## Error

```
Traceback (most recent call last):
  ...
  File ".../cubi_tk/sodar_common.py", line 48, in get_assay_irods_path
    assay, _ = self.get_assay_from_uuid()
  File ".../cubi_tk/sodar_api.py", line 408, in get_assay_from_uuid
    multi_assay_study_warning(assays=assays_)
TypeError: multi_assay_study_warning() got an unexpected keyword argument 'assays'
```

## Steps to reproduce

Any command that goes through `RetrieveSodarCollection` / `get_assay_from_uuid()`
with:

1. `--yes` set,
2. no `--assay-uuid` given, and
3. a project whose study has >=2 assays.

Example: our SODAR project has a study with two assays (panel sequencing and
whole genome sequencing). Running the assay-agnostic path with `--yes` crashes at
the warning call.

## Expected behaviour

With `--yes` and multiple assays, cubi-tk should warn that it will use the first
assay and continue -- not crash.

## Suggested fix

Call the helper positionally, matching its signature:

```python
multi_assay_study_warning(assays_)
```

[...]

PR #345, opened 2026-07-22
fix: bugs pull data and multi assay

### LINKED ISSUE #311: automatic iinit in interactive mode broken
pull data & possibly other functions keeps asking for interactive mode when iinit is triggered, although already in interactive mode. possibly the `self.ask` value is wrongly set.

slaemmle:
closes #311 and #341


Nicolai-vKuegelgen (approved):
LGTM

more of a hotfix, but at least it should make things (more) usable for now

